PR #164, opened 2026-02-18
feat(legend): <NA> gets marker shape updated in settings assignment

### LINKED ISSUE #153: [BUG] <NA> should also get a marker shape update
### Bug Description

When I check the "Include shapes" in the legend settings, the <NA> label should also get a new shape. It should not be handled specially; it should have the same functionalities as any other legend category.

### Code of Conduct

t03i:
Fixes #153

## TLDR

<NA> gets assigned a regular shape now. 
This was explicitly prevented previously.

### DIFF
--- a/docs/guide/data-format.md
+++ b/docs/guide/data-format.md
@@ -53,6 +53,12 @@ The columns `gene_name`, `protein_name`, and `uniprot_kb_id` are **tooltip-only*
 - **Categorical**: Text values (taxonomy, family). Distinct colors per category.
 - **Multi-Label**: Semicolon-separated values (e.g., `EC:1.1.1;EC:2.1.1`). Displayed as pie charts.
 
+### Missing Values
+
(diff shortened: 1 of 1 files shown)


peymanvahidi:
We had this problem that N/A values had 6 different internal representations across the codebase ([], 'null', '', 'N/A', '__NA__', nullishConfiguredColor), causing subtle bugs. So I refactored it

Changes:
  - Normalize at source: DataProcessor and legend.ts now convert null/empty/whitespace to '__NA__' via a shared toInternalValue() in @protspace/utils
  - Simplify downstream: Removed nullishConfiguredColor, length === 0 branches, and fixed computeAllHidden in style-getters
  - Fix export-utils: Replaced 'null' string literals with LEGEND_VALUES.NA_VALUE
  - Fix tooltip: Converts '__NA__' → 'N/A' for display, filters from metadata fields
  - Tests: Added 19 new tests covering toInternalValue, DataProcessor edge cases, and export-utils N/A handling

PR #202, opened 2026-03-24
fix(scatter-plot): extend brush extent to full canvas for edge selection

### LINKED ISSUE #189: [BUG] Selection handle is limited to min-max axis boundaries of 2D space
## Bug Description

When using the selection handle (lasso/rectangle select), the current selection area is constrained to the min-max x and y axis boundaries of the 2D embedding space. This is problematic because:

- Points near the edges of the plot cannot be easily selected
- Users cannot drag the selection rectangle/lasso outside the data range to capture border points
- It forces users to be very precise when selecting points near the boundaries

**Expected behavior:** The selection tool should allow drawing selections anywhere on the canvas, without being limited to the min-max coordinates of the data. This would make it much easier to select points at the edges of the embedding space.

## Code of Conduct

jcoludar:
## Summary

- Expand the D3 brush .extent() from the margin-constrained plot area to the full SVG canvas ([0, 0] to [width, height])
- This allows users to draw selection rectangles beyond the data min/max boundaries, making it easy to capture points near the edges of the embedding space

## Root cause

The brush extent was set to [margin.left, margin.top] to [width - margin.right, height - margin.bottom], which clamped the draggable area to the plot margins. D3 brush automatically constrains selection coordinates to its extent, so edge points could not be reached.

## Verification

- 534 unit tests pass
- 29 Playwright e2e tests pass
- pnpm precommit (format + lint + type-check + knip) clean
- Brush coordinates and scale coordinates remain in the same local coordinate system (the brush group pre-transform space), so zoom interactions are unaffected

Fixes #189

### DIFF
--- a/packages/core/src/components/scatter-plot/scatter-plot.ts
+++ b/packages/core/src/components/scatter-plot/scatter-plot.ts
@@ -746,8 +746,8 @@ export class ProtspaceScatterplot extends LitElement {
         // Keep the UX simple: no resize handles, just drag a rectangle.
         .handleSize(0)
         .extent([
-          [config.margin.left, config.margin.top],
-          [config.width - config.margin.right, config.height - config.margin.bottom],
+          [0, 0],
+          [config.width, config.height],
         ])
         .on('end', (event) => this._handleBrushEnd(event));
 


tsenoner (commented):
This extends the selection area a bit. However, when zooming out and then selecting (btw, we don't have any lasso tool as you describe in the summary) the area is still constrained to the middle.
Please revisit this.

jcoludar:
Superseded by #213 which uses viewport-tracking approach instead.

PR #213, opened 2026-03-31
feat(scatter-plot): fix brush viewport extent and add lasso selection (#189, #208)

### LINKED ISSUE #189: [BUG] Selection handle is limited to min-max axis boundaries of 2D space

### LINKED ISSUE #208: feat: add lasso selection mode
## Feature Request

Add a **lasso selection** mode as an alternative to the existing rectangle selection. When the user activates the selection tool, rectangle selection should remain the default. An option to switch to lasso selection mode should be available, allowing users to draw a freeform outline around points. Selected points should behave the same regardless of which selection method was used.

jcoludar:
## Summary
- Fix brush selection dead-zones at canvas edges by computing viewport-aware extent via inverse zoom transform
- Replace O(n) linear scan in `_handleBrushEnd` with `QuadtreeIndex.queryByPixels()` for spatial pruning
- Fix RAF + setTimeout race condition by consolidating into nested requestAnimationFrame calls
- Add lasso (freeform) selection as alternative to rectangle brush, with inline toggle in control bar
- Add `queryByPolygon()` to QuadtreeIndex using AABB pruning + ray-casting point-in-polygon
- Extract shared `_commitSelection()` to DRY up brush and lasso handlers
- Allow scroll-wheel zoom and double-click reset while in selection mode (drag still draws selection)
- Fix selection rendering: two-pass WebGL (unselected blend-OFF for flat fading, selected blend-ON for correct MSAA anti-aliasing), skip edge darkening on faded points
- Add brush/lasso cleanup in `disconnectedCallback` to prevent listener leaks
- Replace `waitForTimeout` anti-pattern with `waitForFunction` in E2E tests
- Extract shared Playwright helpers into `app/tests/helpers.ts`
- Document lasso selection in scatterplot guide, control bar docs, API reference, and product tour

Closes #189
Closes #208


tsenoner:
@peymanvahidi please have a look at the final implementation. Ivan improved the selection area. I went over it and used QuadtreeIndex and also implemented lasso selection.

peymanvahidi:
### Code review

Found 1 issue:

1. `_updateBrushExtent()` is called on every scroll-wheel zoom event while in rectangle selection mode (line 882). This method calls `this._brushGroup.call(this._brush)` (line 1073), which re-applies the D3 brush to the group. D3's brush re-application during an active drag gesture resets its internal drag state, silently dropping any in-progress rectangle selection. Since commit `5fc4e89` explicitly keeps scroll-wheel zoom active during selection mode (only `mousedown.zoom`/`touchstart.zoom`/`touchmove.zoom`/`touchend.zoom` are disabled), a user who scrolls while dragging a brush rectangle will lose their in-progress selection without feedback. Consider updating the brush extent only on `zoom end` events rather than on every `zoom` tick, or guard against re-applying the brush while D3's brush drag is active.

https://github.com/tsenoner/protspace_web/blob/5fc4e89fa3c49f631c9cf9b00b65d0d8777d86ea/packages/core/src/components/scatter-plot/scatter-plot.ts#L880-L883

https://github.com/tsenoner/protspace_web/blob/5fc4e89fa3c49f631c9cf9b00b65d0d8777d86ea/packages/core/src/components/scatter-plot/scatter-plot.ts#L1058-L1074

peymanvahidi (approved):

Issue #101, opened 2025-12-19
[BUG] Duplicate Badge only available in the PCA


peymanvahidi:
### Bug Description

Duplicate badges are not visible in the umap, even if you turn this option on.

### Steps to Reproduce

Turn on the Duplicate Badges in the legend setting. (in the PCA)
You see the badges.
Now change it to the UMAP.
You don't see them.

### Relevant log output

```shell

```

### Expected Behavior

I expect to have them in all the available projections we have.

### System Information

macos, chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #223, opened 2026-04-07
Enhancements: Z-order issue

### LINKED ISSUE #121: [BUG] duplication badges lagging behind
### Bug Description

Duplication badges are lagging or wrongly displayed when hiding legend categories and when switching between dimensionality reduction methods.

### Steps to Reproduce

1. enable duplication badges
2. hide legend items
3. switch dim reduction methods back and forth

### Relevant log output

```shell

```

### Expected Behavior

Duplication badges get updated according to the displayed dots in the scatterplot/canvas.

### System Information

macos - chrome

### Additional Context

_No response_

### Code of Conduct

### LINKED ISSUE #169: [BUG] Colorization behaves unexpected when changing ordering (decreasing and increasing)
### Bug Description

1. Load the homsa_drosop dataset
2. Select the species annotation
3. Revers Z-order
4. Colors invert in canvas

### Code of Conduct

peymanvahidi:
This PR fixes the issues below:
- fixes #169
- fixes #121
- addresses #101 (see note)

## Notes

**#101 status.** Duplicate badges are now grouped per the *current* projection: two points stack iff they share exact coords in the displayed view. PCA collapses identical embeddings to the same coordinates, so badges appear there. UMAP/t-SNE generally don't co-locate them, so badges don't appear under those projections.

**`html { font-size: clamp(0.75rem, 0.87vw, 1.125rem) }` in `app/src/index.css`.** Fluid base font size, 12px to 18px between ~1379–2069px viewport width. Intent: improve readability on wider displays. Tested on standard desktop viewports (~1440–1920px, MacBook 14"/16" + external monitor).

**ESC cascade.** ESC priority is now: (1) close open duplicate-badge spider, (2) clear selection, (3) toggle off selection mode.

**New renderer API.** `WebGLRenderer.invalidateDepthOrder()` lets callers signal a depth-sort invalidation without invalidating the position cache. Used by scatter-plot's z-order/color-mapping handlers.


peymanvahidi:
next: add docs for the changes (especially badges)

peymanvahidi:
@tsenoner ready to merge

peymanvahidi:
check the docs page, Using the Explore Page -> Navigating the Scatterplot -> Duplicate Points for the gif

tsenoner:
Hi @peymanvahidi,
I noticed some odd behavior when clicking a duplicate badge.

**Bug:** Only the top-most point of the duplicate group renders at the expected position — the other duplicate points appear elsewhere on the plot. This may be a UMAP-specific issue where overlapping/duplicate points aren't rendered at the same coordinates. Could you investigate and propose a fix? See the screen recording below:

https://github.com/user-attachments/assets/2a1ea64c-c4f1-4726-a8b6-64e69f7dac66

**Additional request:** ESC should also close an open duplicate badge, mirroring the behavior we have for selections. Proposed precedence:
1. First ESC → close the open duplicate badge (if any)
2. Next ESC → clear the selection

packages/core/src/components/scatter-plot/scatter-plot.ts
```diff
@@ -617,6 +634,76 @@ export class ProtspaceScatterplot extends LitElement {
     this._invalidateVirtualizationCache();
   }
 
+  /**
```
tsenoner: Why are we doing it this way now? This might be the issue I observed in the screen recording above; under UMAP, members of a cross-projection group don't share canvas coordinates, so the spider opens at the clicked anchor but the other members stay rendered as separate dots elsewhere on the plot.
Reverting to per-projection grouping should fix both.

Please keep the rest of the duplicate-badge fixes (the `_cancelDuplicateStackCompute` guard and the `_scheduleDuplicateOverlayUpdate(true)` after quadtree rebuild) — those address the actual #121 lag. Can you verify #121 still reproduces without the union-find layer?

tsenoner:
A few things still missing before this is ready:

- **ESC handler** (per my earlier comment): close badge → then clear selection.
- **Unit test** for the duplicate-stack logic — at minimum a Vitest covering legend hide and projection switch, so #121 doesn't regress.
- **PR title typo:** "Ehnacements" → "Enhancements".
- **`html { font-size: clamp(...) }`** in `app/src/index.css` — please add a one-liner to the PR description explaining the intent and the viewport range you tested, since it affects every `rem`-sized element in the app.

packages/core/src/components/scatter-plot/scatter-plot.ts
```diff
@@ -1705,9 +1812,23 @@ export class ProtspaceScatterplot extends LitElement {
...
     this._expandedDuplicateStackKey =
       this._expandedDuplicateStackKey === stackKey ? null : stackKey;
+
+    // When opening a cross-projection duplicate stack, re-anchor the stack
+    // position to the clicked point so the spiderfy opens on top of it
+    // rather than on whichever group member happened to be iterated first.
+    if (this._expandedDuplicateStackKey && anchorPoint) {
+      const stack = this._duplicateStackByKey.get(stackKey);
+      if (stack && this._scales) {
+        stack.x = anchorPoint.x;
```
tsenoner: The re-anchor mutates `stack.x/y/px/py`, but `_duplicateStackByKey` is rebuilt on every viewport recompute, so the anchor reverts as soon as the user pans or zooms while the spider is open. Store the anchor separately (e.g. `_expandedSpiderAnchor`) and re-apply it on each rebuild while the spider is open.

packages/core/src/components/scatter-plot/scatter-plot.ts
```diff
@@ -367,6 +372,15 @@ export class ProtspaceScatterplot extends LitElement {
     this._zOrderMapping = customEvent.detail.zOrderMapping;
     // z-order affects GPU depth; force a fresh style getter cache so getDepth sees the new mapping
     this._styleGettersCache = null;
+
+    if (this._plotData.length > 0) {
```
tsenoner, line 376: The comment says the renderer's depth-sampling optimization "can miss swapped z-order between categories." If that's true, it's a renderer bug — let's file a follow-up issue against `webgl-renderer.ts` rather than carrying this position-cache invalidation as a permanent workaround.

packages/core/src/components/scatter-plot/scatter-plot.ts
peymanvahidi: Verified. `_computeCrossProjectionDuplicates` was indeed grouping members that don't share coords in the current projection, which is the UMAP misbehavior you recorded. Removed the union-find and changed the key to per-projection coord identity in `2aca991`. Identical embeddings still stack under PCA; UMAP/t-SNE no longer fabricate stacks for physically-separated points. The `_cancelDuplicateStackCompute` guard and the `_scheduleDuplicateOverlayUpdate(true)` call after quadtree rebuild are preserved (the #121 fix). Locked the new behavior with Vitest in `308099b` covering legend-hide, projection-switch, and the UMAP-jitter case.

packages/core/src/components/scatter-plot/scatter-plot.ts
peymanvahidi: Fixed in `2c50aef`. Anchor is now stored as `_expandedSpiderAnchor: { stackKey, x, y } | null` on the component, set in `_toggleSpiderfy`, and re-applied via `_applyExpandedSpiderAnchor()` inside `_ensureDuplicateStacksForViewport` after every rebuild. Verified by forcing a viewport recompute while the spider is open. The new stack object is a different reference, but the coords match the original anchor.

packages/core/src/components/scatter-plot/scatter-plot.ts
peymanvahidi: Fixed at the renderer in `6be0b82`. Added a separate `depthOrderDirty` flag and a public `WebGLRenderer.invalidateDepthOrder()` that triggers a re-sort without touching the position cache. `_handleZOrderChange` and the non-color-only path of `_handleColorMappingChange` now call `invalidateDepthOrder()`. The position-cache-invalidation workaround and the misleading comment are gone.

peymanvahidi:
Addressed all items from your review:

1. **UMAP duplicate-badge misbehavior and union-find revert**: see thread reply on `scatter-plot.ts:637`. Reverted to per-projection grouping in `2aca991`.
2. **ESC handler**: implemented in `0961fd5`. ESC now closes an open duplicate-badge spider first; the next ESC clears the selection.
3. **Spider anchor persistence on pan/zoom**: see thread reply on `scatter-plot.ts:1825`. Stored separately in `2c50aef`.
4. **Renderer depth-sampling workaround**: see thread reply on `scatter-plot.ts:376`. Replaced with a proper `WebGLRenderer.invalidateDepthOrder()` API in `6be0b82`.
5. **Unit test for duplicate-stack logic**: added in `308099b` covering legend-hide and projection-switch (locks #121 against regression). 15 new tests, total now 824.
6. **PR title typo**: fixed (Ehnacements to Enhancements).
7. **`font-size: clamp(...)` rationale**: added to the PR description.

#101 note: per-projection grouping means UMAP/t-SNE generally don't show duplicate badges (they don't co-locate identical embeddings). PCA still does. Happy to revisit if you'd prefer "badges by embedding identity regardless of projection," but that's the behavior that produced the visual misbehavior you recorded.

tsenoner:
Why are the same sequences in UMAP not overlapping as they do in PCA?
e.g.: `P0C8L1` and `P0C8L1` have the exact same sequences and overlap in PCA but not on UMAP. There is even another entry that is not the same that is in between the two: `P0C185`.
Is that a limitation/property of UMAP?

peymanvahidi:
I think because it's stochastic.

tsenoner:
Yeah, came to the same conclusion. So we just have to keep in mind that, for the UMAP, duplicated badges might not serve any purpose. Which is fine.

tsenoner (approved):

PR #228, opened 2026-04-19
fix(legend): NA-handling redesign + numeric type inference

### LINKED ISSUE #226: [BUG] Floats aren't recognized as numeric and are still shown as categoric values
### Bug Description

An annotation should be shown as continuous if it is either an integer or a float; at the moment, float isn't working.

### Code of Conduct

### LINKED ISSUE #230: [BUG] N/A sorts to top of numeric legend after settings reset
## Description

When changing the numeric bin count in legend settings and then hitting Reset, the N/A entry appears at the top of the legend instead of the bottom.

## Steps to reproduce

1. Load a dataset with missing values in a numeric annotation (e.g., `edge_few_nan_many_numbers` from `data_custom.parquetbundle`)
2. Open legend settings, change bins to 5, save
3. Hit Reset
4. N/A appears at the top of the legend instead of the bottom

## Root cause

Lit's `updated()` lifecycle fires `_rebuildLegendItems()` when `maxVisibleValues` changes during reset, **before** `syncNumericAnnotationSettings()` triggers re-materialization with new bin IDs. This intermediate rebuild creates legend items from stale annotation data. When re-materialization completes and triggers a second rebuild, the stale `visibleValues` set puts `__NA__` (the only surviving value from the old bins) first as a "visible" entry, with all new bins appended after as "non-visible".

## Expected behavior

N/A should always appear at the bottom of the numeric legend, regardless of prior settings changes.

## Notes

The `sortAndLimitItems` sort function correctly sorts N/A last in all modes. The issue is upstream in the `workingVisible` path of `_updateLegendItems`, which prioritizes previously-visible values over newly-appearing ones.

FlorinSenoner:
Closes #226, #230. Continues the numeric type-inference fix with a full redesign of NA handling so categorical NA behaves like any other category and numeric NA stays consistently locked.

## What changed

**NA pipeline**
- Single source of truth: new `@protspace/utils/missing-values` module (`MISSING_VALUE_TOKENS`, `NA_VALUE`, `NA_DISPLAY`, `NA_DEFAULT_COLOR`, `normalizeMissingValue`).
- Boundary normalization at row read in `conversion.ts`: every spelling of missingness collapses to JS `null` upfront. Downstream code now relies on `value == null`.
- Categorical columns with mixed missing forms (`null`, `""`, `"NA"`, `"N/A"`, `"NaN"`, `"None"`, `"null"`, `"missing"`, case-insensitive) produce a single canonical `__NA__` legend row covering all those entries — fixes the prior "string NA leaks as its own category" bug.
- Categorical NA participates in sort like any value (manual / alpha / size). Falls into "Other" when `maxVisibleValues` is exceeded. Default color `#DDDDDD`, override-able via the legend customizer.
- Numeric NA pinned to the end of the legend regardless of sort mode. Color and shape locked to `#DDDDDD` + circle. Persistence skips numeric NA writes.

**Legend keyboard reorder Escape (fixes #230)**
- Added a window-level capture-phase handler that cancels active keyboard reorder regardless of focus. The drag-handle's own Escape branch only fires while focus stays on that handle, but Lit re-renders during ArrowDown can briefly move focus to `<body>` before the rAF focus-restore lands. The new handler closes that gap, which also fixes the N/A-at-top issue when resetting numeric bin count on a column with missing values.

**Dead code removal**
- `MISSING_VALUE_MARKERS` and `isMissingValueMarker` (replaced by `normalizeMissingValue`).
- `LEGEND_VALUES.NA_VALUE` / `NA_DISPLAY` / `NA_COLOR` (moved into the new module).
- `SPECIAL_COLORS` map and the NA branch in `visual-encoding.ts`.
- The unconditional NA-pin in `sortFn` (replaced with a gated `isNumericSource` branch).
- The `extractingNA` / `mergingNA` paths in `sortAndLimitItems`.
- The NA-skip on color-conflict resolution.
- The unreachable `slot < 0` non-Other branch.
- All literal `'N/A'` comparisons in production source.

## Breaking changes

- Strings `"-"` and `"."` are no longer treated as missing — they appear as real categorical values. Numeric columns containing them demote to categorical.
- String forms `"Infinity"` / `"-Infinity"` / `"inf"` / `"-inf"` are also no longer in the missing set (numeric `±Infinity` is still treated as missing via `Number.isFinite`). Same demotion behavior.
- Persisted legend customizations under leaked categories like `"NA"` or `"None"` won't migrate — users see modal defaults until they re-customize.


packages/core/src/components/data-loader/utils/conversion.ts
```diff
@@ -48,6 +48,36 @@ interface AnnotationInferenceResult {
...
+ * Compared case-insensitively. These do NOT force a numeric column to categorical.
+ */
+const MISSING_VALUE_MARKERS = new Set([
+  'nan',
+  'na',
+  'n/a',
+  'null',
+  'none',
+  '-',
+  '.',
```
tsenoner: clean up, just using the most basic. 
na, nan, null, none

Issue #200, opened 2026-03-24
[DOCS] Update documentation to reflect latest changes


tsenoner:
### Description

The documentation is out of date in several areas and needs to be refreshed to match the current state of the tool.

### What needs to be updated

- **CLI flag rename**: The [Python CLI guide](https://protspace.app/docs/guide/python-cli.html) still references the old `-f` flag for annotation selection instead of the new `-a` flag.
- **Screenshots and GIFs**: The UI has changed since the last documentation update. All PNGs and GIFs need to be regenerated to reflect the current interface.
- **General review**: Go through the rest of the documentation and identify any other sections that are outdated or inconsistent with recent changes.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

tsenoner:
Resolved by #232 — CLI guide rewritten around `protspace prepare` with `-f` → `-a` fix, all 14 documentation screenshots and GIFs regenerated, data preparation docs updated with TED/Biocentral/LocalMAP.

PR #232, opened 2026-04-26
feat(publish): publication-ready figure editor

tsenoner:
## Summary

- Add full-screen Figure Editor for creating publication-ready figures with live preview
- 15 journal presets (Nature, Science, Cell, PNAS, PLOS) with DPI/mm/px linkage
- Interactive overlay tools: circle (with rotation/resize), arrow, label, zoom insets
- Legend customization: 10 positions, multi-column, free-floating with drag
- State persistence via localStorage and parquetbundle
- View fingerprint detection warns when overlays may be stale after projection change
- Simplify export dropdown: promote Figure Editor as primary, add Quick Export secondary
- DRY refactoring: deduplicate mmToPx/pxToMm, extract shared methods, add slider helper
- 35 new tests for overlay controller and compositor (886 total, all passing)
- Comprehensive documentation page for the Figure Editor
- Update Python CLI guide: rewrite around `protspace prepare`, fix `-f` → `-a` flag, add LocalMAP, annotation groups, embedder models (#200)
- Update data preparation docs with TED, Biocentral sources and LocalMAP (#200)
- Dismiss product tour in Playwright screenshot scripts
- Regenerate all 14 documentation screenshots and GIFs (#200)


jcoludar (commented):
*Updated 2026-04-27 — re-verified each finding against the actual code. Three items from my first pass were wrong; striking them and replacing with a tighter, verified list.*

Architecture is sound. A few real correctness gaps in lifecycle / event handling, plus one quiet behaviour regression that should be acknowledged before merge. Nothing rewrite-level.

## Must-fix

**1. Pointer capture leak — no `pointercancel` handler, no `releasePointerCapture`**
`packages/core/src/components/publish/publish-overlay-controller.ts:476` calls `setPointerCapture(e.pointerId)` on `pointerdown`. The file has zero `releasePointerCapture` calls and no `pointercancel` listener. `destroy()` (line 91) only removes pointerdown/move/up. If the OS cancels the pointer (touch interruption, devtools opened mid-drag, browser scroll/zoom takeover), `this.drag.active` stays `true` and the next `pointermove` keeps mutating the selection. Add a `pointercancel` handler that resets `drag`, `handleMode`, `legendDragging`, plus an explicit `releasePointerCapture(e.pointerId)` in both `onPointerUp` and the new `onPointerCancel`. Add `removeEventListener('pointercancel', …)` to `destroy()`.

**2. `destroy()` doesn't abort an in-flight drag**
`destroy()` only removes listeners. It doesn't reset `this.drag.active`, doesn't release pointer capture, doesn't clear `handleMode`/`legendDragging`. `publish-modal.ts:138` and `:466` call `destroy()` from `disconnectedCallback` and `_applyStateAndRebuild`. If destroy fires mid-drag (rapid open/close, preset rebuild while dragging), pointer capture is still held by the now-removed controller; the new controller can't get pointer events for that `pointerId` until the OS releases it. Make `destroy()` symmetric — release capture, clear drag state, then remove listeners.

## Should-fix

**3. Behaviour regression — per-annotation export prefs lost**
The deletion of `export-persistence-controller.ts` (-55), `control-bar-helpers.ts` (-40), and the per-annotation `exportImageWidth/Height/LegendWidthPercent/LegendFontSizePx/LockAspectRatio/IncludeExportOptions/IncludeLegend` state (`control-bar.ts` -432) is replaced by a single global `localStorage['protspace_publishState']` key (`app/src/explore/export-handler.ts:49`). Quick-export now always uses `EXPORT_DEFAULTS`. Existing users with saved per-annotation export preferences will silently lose them on first run. Either keep per-annotation persistence on the Quick Export path (the two storage models can coexist), write a one-time migration that seeds the new global state from the old per-annotation keys, or call this out in release notes. Probably worth confirming this trade-off was deliberate.

**4. `savedPublishState` blindly cast to `PublishState` — `publish-modal.ts:119`**
```ts
this._state = { ...defaults, ...this.savedPublishState } as PublishState;
```
The property type is `Record<string, unknown>` and the validator in `packages/utils/src/parquet/settings-validation.ts:170+` only checks "is object". Corrupted or pre-migration saved state, or a hand-edited parquet bundle, will silently produce broken overlays / out-of-range coords. Add a runtime validator that drops unknown overlay types and coords outside `[0,1]` rather than merging garbage into state.

**5. `toNorm` clamps to `[0,1]` even during handle drags — `publish-overlay-controller.ts:106-107`**
```ts
nx: Math.max(0, Math.min(1, (px - pr.x) / pr.w)),
ny: Math.max(0, Math.min(1, (py - pr.y) / pr.h)),
```
When the user drags an overlay handle past the plot edge, the coord snaps to the boundary. Combined with `applyHandleDrag` using the clamped value, a fast drag past the edge can collapse `rx`/`ry` unexpectedly. Make the clamping opt-in (clamp on creation only, not during handle drag).

**6. No canvas-area cap on `widthPx * heightPx * boost²`**
`publish-modal.ts:699` allows up to 8192×8192. `publish-modal.ts:271` then multiplies by `boost`. Browsers cap canvas at ~16384² or ~268M total pixels — beyond that, `getContext('2d')` returns null or draws to zero-size silently. Pick a `MAX_CANVAS_PIXELS` constant and clamp `boost` against it, ideally surfacing a tooltip when boost is being clamped down.

**7. Light-DOM `document.querySelector('protspace-legend')` — `publish-modal.ts:68`**
Works today because the legend lives in the same document. Breaks the moment the modal is mounted inside another shadow root (embedded mode, third-party integration). Pass legend export state in via property the way `plotElement` is.

**8. `document.fonts.ready` not awaited before compositor `fillText`**
`publish-compositor.ts:188, 272`. Currently fine because only `Arial, sans-serif` is used, but if a future preset specifies a webfont this will silently fall back. Add `await document.fonts.ready` before export rendering.

## Style consistency

`publish-modal.ts:86` only includes `[tokens, buttonMixin, publishModalStyles]` — skips `inputMixin` and `overlayMixins` that the rest of the codebase uses for modals/inputs. The custom 510-line `publish-modal.styles.ts` ends up reimplementing input/overlay scaffolding that already exists. Probably 30 min of refactor to inherit and prune. Naming, lifecycle hooks, and test patterns otherwise match the existing house style.

## Test coverage gaps

Tested well: `dimension-utils` (full math), `journal-presets` (full table), `publish-compositor` `computeLayout`/`computeInsetBoost`/`capturePlotCanvas`, `publish-modal-helpers`, `settings-validation` publish-state branch, overlay creation primitives, basic hit-testing, basic drag, default state, free legend.

Not tested:
- **Lifecycle:** no test that destroy mid-drag releases pointer capture or clears `drag.active`. No `pointercancel` test (because no handler — see #1).
- **Modal element:** `publish-modal.ts` (1080 lines) has zero element-level tests — no render, no `_setupOverlay` reentrance, no fingerprint-warning behaviour, no preset-rebuild path (`_applyStateAndRebuild`).
- **Compositor rendering:** `composeFigure`, `renderInset`, `drawCircleOverlay/drawArrowOverlay/drawLabelOverlay` are uncovered; `publish-compositor.test.ts` only exercises pure layout.
- **Hit-testing edges:** rotated-circle hit-test, label rotation handle, inset-handle drag with constrained aspect, overlay clamping at plot edges.
- **State migration / corrupt save:** no test that bogus `savedPublishState` doesn't crash the modal — see #4.
- **`exportCanvasAsPdf`:** zero tests.

---

**Retracted from my first pass:** the earlier review claimed (a) the inset-boost path double-captures when boost===1 — wrong, line 401 explicitly guards `if (boost > 1)`; (b) the backdrop's `@mousedown=${this._onOverlayClick}` is a bug — overstated, the inner container has `stopPropagation` so canvas drags don't bubble, this is `@click` vs `@mousedown` preference not a bug; (c) dead code at lines 217-223 of the overlay controller — wrong, those lines are real `hitTestCircleHandles` code. My apologies for the noise.

Happy to pair on patches for #1, #2, #4 — they're all small.

jcoludar:
Followup PR opened at #235 against this branch (`feat/publish-editor`), addressing the must-fix and should-fix items from my review above. Targeting your branch directly so it folds into this PR's diff when you take it.

Three items remain deferred there as **Follow-up considerations** rather than patches — design calls (no current users affected) where input from you would be useful, particularly the per-annotation export prefs consolidation: worth confirming whether that was deliberate.

PR #244, opened 2026-05-05
docs: small accuracy fixes (projection list, AlphaFold path, dev port)

tsenoner:
Three small corrections found while auditing docs against current code:

- `docs/explore/structures.md` — clarified that the AlphaFold structure file is fetched directly from the AlphaFold Database API; 3D-Beacons is only used for the model page link.
- `docs/index.md` and `docs/explore/index.md` — projection list updated from PCA/UMAP/t-SNE to include PaCMAP and MDS (already shipped, just undocumented).
- `docs/developers/installation.md` — dev port `5173` → `8080` to match `config/urls.ts`.

### DIFF
--- a/docs/developers/installation.md
+++ b/docs/developers/installation.md
@@ -70,7 +70,7 @@ pnpm install
 pnpm dev
 ```
 
-The dev server runs at `http://localhost:5173`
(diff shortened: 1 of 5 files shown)


tsenoner (commented):
LGTM

PR #245, opened 2026-05-05
docs(tour): clarify Esc, Filter query builder, and Cmd/Ctrl+Click

tsenoner:
Three small tour-copy fixes from the docs/tour audit:

- **Step 5 (Selection Tools)** — Esc has two-stage behavior (clears selection on first press, exits selection mode on second). Tour now describes both stages.
- **Step 6 (Filter, Export & Import)** — Filter is a query builder with AND/OR logic over annotation conditions; the previous "narrows the dataset by annotation values" was an oversimplification.
- **Step 7 (Interactive Scatterplot)** — adds discoverability for Cmd/Ctrl+Click to toggle a single point in or out of the current selection. Uses the existing `isMacOrIos` helper to render the platform-correct modifier (⌘ on Mac/iOS, Ctrl elsewhere).

### DIFF
--- a/app/src/tour/product-tour.ts
+++ b/app/src/tour/product-tour.ts
@@ -215,14 +215,14 @@ const steps: DriveStep[] = [
   shadowStep(CONTROL_BAR, 'selection', {
     title: 'Selection Tools',
     description:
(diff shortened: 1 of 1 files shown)


tsenoner (commented):
LGTM

PR #248, opened 2026-05-06
Feat/fasta prep backend

### LINKED ISSUE #236: [FEATURE] Installing the github protspace setup on one of our clusters
### Feature Request

Installing the github protspace setup on one of our clusters

### Code of Conduct

t03i:
## TLDR

MVP backend implementation. 

## Description

Architecture
```txt
[Browser]
   │ POST /api/prepare (multipart FASTA)
   │ GET  /api/prepare/:id/events  (SSE)
   │ GET  /api/prepare/:id/bundle  (download)
   ▼
[Caddy reverse proxy] ── existing TLS termination, SSE-friendly by default
   │
   ▼
[protspace-prep container]
   │ FastAPI app (uvicorn, single process)
   │ ├─ in-memory job registry: dict[job_id → JobState]
   │ ├─ asyncio.Semaphore(5) bounding concurrent active jobs
   │ ├─ asyncio task per job; pipeline runs via asyncio.to_thread
   │ └─ /var/lib/protspace-prep/jobs/<job_id>/ for bundle artifacts
   │
   ▼
[Biocentral API] (HTTPS, anonymous today)
```
SLO:
- 5 concurrent users supported
- resutls of 1k seqs in ~1 min 

Closes: #236


t03i:
1 min is currently absolutely unrealistic. For 840 sequences, we're breaching 3:30 likely because of Biocentral. Some more profiling is required.

t03i:
Changed e2e to only run nightly/with label. 
Changes regarding colab and progress already in now @tsenoner.

tsenoner:
## Review notes                                                                     
                                                                                                                
  **Likely bugs**                                                                                               
  - `app/src/explore/runtime.ts:140` — the `embedding`/`annotating` branch hardcodes `'Embedding sequences (~3  
  min)…'`, overriding the dynamic `embeddingLabel` from `formatEmbeddingLabel(seqCount)`. A 50-seq job will show
   "~3 min" instead of "~20 sec". Should reuse `embeddingLabel` like the `queued` branch does.                  
  - `api.py:134` — `mark_consumed(job_id)` runs before the response body is sent; the bundle deletion is a      
  `BackgroundTask`. If the SPA's download fetch fails mid-stream, `consumed=True` *and* the file is gone, so    
  retry returns 410. Consider deferring `mark_consumed` until the BackgroundTask completes successfully.
                                                                                                                
  **Colab fallback UX**                                                                                         
  - We want the Colab notebook
  (https://colab.research.google.com/github/tsenoner/protspace/blob/main/notebooks/ProtSpace_Preparation.ipynb) 
  to be a visible alternative — not only during the prep flow as the "Got a larger dataset?" note. Consider   
  surfacing it on the explore page itself (e.g. near the FASTA upload affordance) so users know they can prepare
   the data themselves.                                                                                       
  - When Biocentral is unavailable, the backend already returns `BIOCENTRAL_UNAVAILABLE` with a friendly message
   in `_BIOCENTRAL_FRIENDLY_MESSAGE`, but the message is plain text — the frontend currently doesn't render it  
  as a clickable link to the Colab notebook. When the SPA receives an error with `code === 
  'BIOCENTRAL_UNAVAILABLE'`, it should show a dedicated banner/notification with the Colab link and an          
  explanation that the user can run the same pipeline there. Right now the user just sees an error toast.     

  **Dead / inconsistent config**
  - `Settings.biocentral_endpoint` (`config.py:19`, `PREP_BIOCENTRAL_ENDPOINT`) is loaded but never read. Either
   pipe it into the `protspace embed` subprocess or drop the field.                                             
  - `PREP_SEQUENCE_MIN_COUNT` defaults to `1` in `config.py`, is overridden to `20` in `docker-compose.yml`, and
   isn't mentioned in `services/protspace-prep/README.md`. Pick one and document it.                            
                                                                                                              
  **Worth a TODO, not a blocker**                                                                               
  - In-memory `JobRegistry` — container restart orphans live jobs. Job dirs survive on the volume but `_jobs` is
   empty, so `/bundle` returns 404 until the sweeper runs. Worth flagging in the README and deciding how the SPA
   should react.                                                                                              
  - `_classify_failure` is only applied inside the embed/annotate `TaskGroup` catch (`pipeline.py:115`). A      
  failure in `project` or `bundle` won't get the `BIOCENTRAL_UNAVAILABLE` reclassification. Not a real risk     
  today, but a sharp edge if those steps ever go network-bound.
                                                                                                                
  **Smaller**                                                                                                 
  - `_looks_like_nucleotide` is `all(ch in ACGTUN)` over ≥50 residues — theoretically can false-positive on
  pathological sequences (e.g. poly-A peptides). Consider a ratio-based check if false positives ever appear.

t03i:
- mark_consumed before stream (api.py): fixed. Both consume + unlink now run in a post-stream BackgroundTask, so a failed mid-stream download stays retryable instead of 410. (Slight trade: a narrow window where two concurrent GETs could both stream — fine for this single-user flow.)
- Dead biocentral_endpoint: dropped the field; it had zero readers. Will re-add with an actual consumer if/when the embed step needs it.
- PREP_SEQUENCE_MIN_COUNT: aligned the code default to the deployed 20 and documented it in the README. Tests that exercise other gates pin min=1.
- In-memory JobRegistry restart: added a "Known limitations" note in the README.

## Deliberately not in this PR (filing follow-ups):

- Colab surfacing on explore + BIOCENTRAL_UNAVAILABLE banner: this can be follow up. Toast works for now the new UX would be additional effort (not well suited for this PR)
- _classify_failure scope and nucleotide ratio check: left for future as non blockers.

PR #254, opened 2026-05-13
feat: multi-annotation hover tooltip (#234)

### LINKED ISSUE #234: [FEATURE] Multi annotation tooltip
### Feature Request

When exploring embedding clusters using a specific annotation, showing other annotations in the Hover-Tooltip would be helpful to avoid swapping around annotations a lot. 
As there are potentially a lot of annotations available a manual selection for those tooltip annotations might be the best compromise.

### Code of Conduct

Moomboh:
## Summary
Closes #234.

Lets users opt extra annotation features into the hover tooltip alongside the primary (legend-defining) annotation.

### Dropdown UI (`protspace-annotation-select`)
- Filled **primary indicator dot** on the currently selected row.
- Small **(i) toggle button** on every non-primary row. Click flips that annotation in/out of the hover tooltip; native `title` provides the help text ("Show in hover tooltip" / "Hide from hover tooltip"). `aria-pressed` reflects state.
- Toggle dispatches `tooltip-annotation-toggle` (separate from `annotation-select`); the row label still selects the primary.

### Hover tooltip
- `TooltipView` now carries an ordered `blocks[]` array. Each `AnnotationBlock` has its own `key`, `displayValues`, `numericValue`/`numericType`, `scores`, `evidence`.
- The primary annotation renders first, then any extras in dropdown-add order — each gets the full treatment (header, scores/evidence column, raw numeric row).

### State + persistence
- New `?tooltip=foo,bar` URL param (comma-separated). Parser dedupes, drops the primary, drops names not in the dataset, marks anything off for normalization on next URL write.
- Dataset-scoped localStorage (`protspace:tooltip-annotations:<hash>`) seeded on dataset load when the URL does not pin a value; kept in sync afterwards via the existing view-change subscription.
- Promoting an extra annotation to primary drops it from the extras (no double-render).

### Test coverage
- 11 new url-state tests, 5 localStorage-store tests, 6 component tests for the dropdown UI, multi-annotation tests for `buildTooltipView`.
- New Playwright spec `multi-annotation-tooltip.spec.ts` (3 tests) covers the dropdown indicator/toggle, URL writes on toggle, and round-trip removal — all pass locally.
- Existing `818 + 73 + 243` unit/component tests still pass.

## Test plan
- [x] Primary annotation visually distinguished in the dropdown (radio-dot)
- [x] (i) icon toggles tooltip inclusion without changing the primary annotation
- [x] Hover tooltip shows extra blocks for opted-in annotations in stable order
- [x] State persists across reloads (URL + dataset-scoped localStorage)
- [x] Keyboard nav unchanged; `aria-pressed` + `title` on (i) for a11y
- [x] Type-check, knip, unit/component tests, docs build (via pre-commit)
- [x] Local Playwright e2e spec for the new dropdown + URL plumbing

### DIFF
--- a/app/src/explore/control-bar-events.ts
+++ b/app/src/explore/control-bar-events.ts
@@ -40,6 +40,10 @@ export function bindControlBarEvents({
 
   addControlBarListener('projection-change', viewController.handleUserProjectionChange);
 
+  addControlBarListener('tooltip-annotations-change', () => {
+    viewController.handleUserTooltipAnnotationsChange();
+  });
+
   addControlBarListener('selection-disabled-notification', (event: Event) => {
     const customEvent = event as CustomEvent<SelectionDisabledNotificationDetail>;
     notify.warning(getSelectionDisabledNotification(customEvent.detail));
--- a/app/src/explore/data-renderer.ts
+++ b/app/src/explore/data-renderer.ts
@@ -25,6 +25,7 @@ interface ResolvedInitialView {
   annotation: string;
   projectionIndex: number;
   projectionName: string;
+  tooltip: string[];
 }
 
 function resolveRenderableView(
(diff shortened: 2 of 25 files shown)


tsenoner:
- Ensure that the tooltip uses the available vertical space and does not go outside the bottom if selecting points in the lower area.
- Update the annotation selection toggle icon to be a closed and open eye
- make the numeric value in the tooltip non-bold and simply call it "value" not "raw value"

Moomboh:
Addressed the three review items:

- **(i) → eye icons** — the per-row tooltip toggle now uses an open-eye when the annotation is included in the hover tooltip and a closed-eye (slashed) when it isn't (`packages/core/src/components/control-bar/annotation-select.ts`).
- **"Raw value:" → "Value:" (non-bold)** — renamed the label and replaced the bold `.label` styling with new `.tooltip-annotation-raw-label` / `.tooltip-annotation-raw-value` classes (both `font-weight: normal`, muted color) (`packages/core/src/components/scatter-plot/protein-tooltip.{ts,styles.ts}`). The numeric-binning Playwright test that greps the row was updated accordingly.
- **Tooltip overflow at the bottom** — `scatter-plot.ts` now measures the real rendered tooltip height in `updated()` and uses it for the vertical clamp in `_getTooltipStyle()` instead of a hardcoded `tooltipApproxHeight = 160`, so multi-block tooltips on points near the bottom edge stay fully inside the plot viewport.

peymanvahidi (approved):

PR #259, opened 2026-05-14
refactor: control bar cleanup and filtering improvements

### LINKED ISSUE #196: [REFACTOR] Remove 3D projection plane selector and simplify to 2D-only rendering
## Problem

The web frontend supports 3D projections from the Python backend, but there is
**no true 3D rendering**. When a 3D projection is loaded, a "Plane" dropdown
(XY / XZ / YZ) appears that picks which two of three coordinates to plot on the
2D scatter plot. Users cannot rotate or interact with the data in 3D space —
it is purely orthographic 2D slicing.

This adds conditional `projectionPlane` plumbing across the control bar, scatter
plot, and data processor for minimal analytical value.

## Proposal

Remove the plane selector and all 3D-specific logic from the web frontend.
3D projections from the Python backend will still load — the `z` coordinate is
simply ignored, and only `x` / `y` are rendered.

The Python CLI (`protspace` package) is not affected and continues to support
3D projections independently.

## Code of Conduct

### LINKED ISSUE #227: [FEATURE] Different filtering for numeric values and categorical values
### Feature Request

Numerical values should support filtering for larger-than, smaller-than, or an interval.

### Code of Conduct

### LINKED ISSUE #256: [BUG] Filtering logic of selecting AND for multilabel is not working
### Bug Description

When selecting a multilabel with the filtering logic and selecting them one by condition, separated by AND, it should only select the multilabel once that has both conditions. Currently, this doesn't work and shows/computes 0 matching points.

### Code of Conduct

### LINKED ISSUE #257: [BUG] Applying the filtering logic twice represents the filtered points with an increased transperency.
### Bug Description

Filtered points shouldn't be more transparent, not even when applying two filterings after each other.

### Code of Conduct

peymanvahidi:
- [X] Closes #196

Removes the 3D projection "Plane" (XY/XZ/YZ) dropdown and all `projectionPlane` plumbing from the web frontend. 3D Parquet files continue to load, `z` is silently discarded and only `x` / `y` are rendered. The Python `protspace` CLI is unaffected.

Changes:

1. **control-bar:** remove plane selector UI, `handlePlaneChange`, `projection-plane-change` event, the `projectionPlane` property, the cross-element assignment that pushed plane onto the scatter-plot, and `projectionPlane?` from `ScatterplotElementLike`.
2. **scatter-plot:** remove `projectionPlane` property, xz/yz coord remap, `point.z` write, and `projectionPlane` from change-detection gates. `DataProcessor.processVisualizationData` is now called with four args.
3. **data-processor:** remove the `projectionPlane` parameter and the xz/yz branch. `PlotDataPoint` loses its optional `z?` field. Tests updated: xz/yz mapping tests removed, "preserves z" test replaced with "drops the z coordinate for 3D projections."
4. **control-bar helpers:** delete the now-unused `isProjection3D` and `getProjectionPlane` helpers and their tests.
5. **scatter-plot (chore):** drop stale "x/y/z" from the `_updatePlotDataCoordinates` fast-path comment, it now correctly says "x/y" after `PlotDataPoint.z` was removed in commit 3.

---

- [x] Closes #227
- [x] Closes #256 
- [x] Closes #257


tsenoner:
The filtering logic needs a complete revisit.
BUG reproduction:
- select filter for "protein_family": phospholipase A2 family -> apply -> 546 proteins selected
- open the filter panel again without any changes -> apply -> 19 proteins selected
-  open the filter panel again without any changes -> apply -> 19 points are more transparent ( #257 )

peymanvahidi:
@tsenoner I think the issues are fixed

tsenoner:
Hi @peymanvahidi, great progress.
There are still a bunch of bugs in the filtering logic. Please investigate them:
- applying a filter-> open the filter again -> "Reset All" -> "cancel": This will update the legend but not the visualization
- Applying an empty filter without a condition shows as if we have a filter applied in the control bar. Fix this by disallowing "Apply Filter" without a condition. Therefore, it must be closed.
- The behavior of the "Cancel" button is odd in general. When a successful filter is applied, edited, and then canceled, the edited changes are shown when reopening (which is misleading, as they aren't applied). Therefore, I think cancel should always restore the state it was in before (except when pressing "Reset All", when all Filters should be removed + the legend & canvas updated).
- Filtering on multilabel still doesn't work properly. Choose keyword -> When selecting a category, the numbers shown don't represent the numbers actually shown in the legend (see image). When selecting this number, they are filtered by it, not the actual number (e.g., Secreted only displays 41 instead of the 6169 proteins).
- Display the number of proteins that are displayed in the lower left corner, the same way we do when isolating. Ideally, we want to display Y/X proteins. Thinking about it, we always want to display the number of Proteins; so, simply put, we want to always show Y/X proteins. Where Y and X are the same when no filtering or isolation is applied.
- When a filter has been applied, and the filter menu gets reopened, it takes quite a long time for the texts: "1663 of 6183 proteins matched" and "1,663 proteins match" to be displayed. Please investigate the source of the delay and how to improve it. Also, let's be consistent and either show or not to show the thousand separator throughout (I'm in favor of displaying it everywhere).

peymanvahidi:
I fixed first 5 bullet points.
I don't have the issue of the last bullet point, could you investigate why it's happening for you?

PR #272, opened 2026-06-06
feat(annotations): mark predictions and surface per-annotation docs

### LINKED ISSUE #221: [FEATURE] organize predictions annotations
### Feature Request

Coordinate with the backend what the best annotation labels are for predicted annotations and then how to mark them as predictions in the frontend.
Further integrate some information connection to the documentation that describes each of the annotations (connection to the frontend or backend)?

### Code of Conduct

tsenoner:
Enriches the annotation reference docs and refines the in-app annotation picker.

### Docs
- Deep, source-linked detail per annotation (primary-source-verified), kept in a docs-only module so the prose doesn't ship in the app bundle; the brief popover text stays the single source of truth.
- Taxonomy ordered by rank depth (root→species), with the rank ladder in its intro.
- Local dev: the docs port is driven from one constant with `strictPort`, so a silent port bump can't break the `/docs` proxy and the in-app "Learn more" links.

### Predicted annotations
- Mark the de-novo / ML / structure predictors (Biocentral, Phobius `signal_peptide`, TED `ted_domains`) with the ⚡ Predicted badge; reference signature databases (Pfam, CATH, …) and curated/factual data (UniProt, Taxonomy) stay unmarked. Provenance traced against the `protspace` Python backend.
- The dropdown groups strictly by source and shows a per-row ⚡ badge (decoupled from grouping, so marking InterPro/TED no longer empties their groups).

### Annotation picker UX
- The info popover opens on hover and keyboard focus and floats beside the dropdown panel (arrow pointing at the row), so every row name stays visible while scanning the ⓘ column; it stays hoverable for "Learn more".
- The ⓘ is now a leading icon, next to where the popover appears (short pointer travel).
- The dropdown menu fits its content instead of stretching to the full trigger width.


tsenoner:
Add a bit more information in the docs and confiogure the docs properly to work on local (configure it properly - 5175 vs 8000)

peymanvahidi (approved):
LGTM

Issue #271, opened 2026-06-06
Regenerate demo bundles: replace stale length_fixed/length_quantile with numeric length


tsenoner:
## Problem

The demo datasets in `app/public/data/*.parquetbundle` still carry the legacy pre-binned columns `length_fixed` and `length_quantile` (string bin ranges like `50-100`, `200-400`). These were removed in protspace **v4.0.0** in favour of a single raw numeric `length` that the frontend bins on the fly (fixed/quantile strategies in the legend).

The current default bundle (`app/public/data.parquetbundle`) was already regenerated and uses `length`. The other selectable demo datasets were not, so they still show the stale columns — which now surface in the annotation dropdown under "Other" (prettified as "Length fixed" / "Length quantile") with no description.

Affected bundles (all under `app/public/data/`):
`5K`, `40K`, `35K_ec_brenda`, `7K_toxprot`, `105K_homoSapiens_drosophilaMelanogaster`, `127K_beta_lactamase`, `573K_swissprot`, `beta_lactamase_ec`, `beta_lactamase_pn`, `phosphatase`.

## Why it can't be fixed in place

The bundles store only the bin-range strings, not the raw lengths, so a correct numeric `length` cannot be reconstructed by editing the parquet — the source pipeline must regenerate it.

## Proposed fix

Regenerate each demo bundle through the protspace pipeline so it emits a single numeric `length` (and drops `length_fixed`/`length_quantile`), mirroring the toxprot demo regeneration (`protspace/scripts/generate_toxprot_demo.py`, design: `protspace/docs/superpowers/specs/2026-04-30-toxprot-demo-regeneration-design.md`). Requires the source `.h5` embeddings + annotation fetches per dataset.

## Context

Surfaced while implementing the predicted-annotations / annotation-docs work (#221). The annotation-metadata registry already maps `length` → numeric "Sequence length"; unknown columns degrade gracefully, which is why the stale columns currently appear under "Other".

PR #273, opened 2026-06-06
chore(data): restore default bundle and regenerate phosphatase demo

tsenoner:
Two demo-data updates:

- **Restore default bundle** — reverts the temporary toxprot-2025 showcase default (5792fa4) back to the regenerated ToxProt 2025 export with numeric `length` (948bbd3), now that the showcase window ended.
- **Regenerate phosphatase demo** — rebuilds `app/public/data/phosphatase.parquetbundle` via the protspace backend (uv-locked env) from the phosphatase FASTA with `-a all` (1587 proteins, esm2_650m, pca2+umap2). Replaces the legacy pre-binned `length_fixed`/`length_quantile` with numeric `length` and adds the full annotation set incl. the four `predicted_*` columns. Doubles as an end-to-end check of the new lockfile.

Partially addresses #271 (phosphatase bundle).


tsenoner:
Superseded by #272 — these two demo-data commits were folded into the #221 branch so everything merges as one PR.

Issue #249, opened 2026-05-06
[FEATURE] Look into the limitations of E2E and why it take so long


tsenoner:
### Feature Request

How can we optimize the E2E testing?
Do we need all the tests?
Can we remove some of them?

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #274, opened 2026-06-08
Fix and speed up the e2e suite (numeric-binning query builder)

### LINKED ISSUE #237: [BUG] numeric-binning Playwright tests broken after filter-menu → query-builder rewrite
## Bug Description

After running the full `numeric-binning` Playwright project (`pnpm test:e2e --project=numeric-binning`), 14 tests fail. They split into three clusters, all of which look like **test-side mismatches with shipped UI changes**, not actual UI regressions — the filter button + menu and the demo dataset both work correctly in the app at `localhost:8080`.

### Cluster A — `.filter-menu` selectors (9 failures)

The test helpers in `app/tests/numeric-binning.spec.ts:460-540` (`openFilterValueMenu`, `setFilterValues`, `disableFilter`, `applyFiltersFromMenu`, `countOpenFilterValues`, `readOpenFilterLabels`) drive a UI that no longer exists.

- **Old DOM the helpers expect:**
  - `.filter-menu` container in `protspace-control-bar`
  - `.filter-menu-list-item` per-annotation rows
  - `.filter-item-checkbox` toggles, `.filter-item-values-button` per row
  - Per-row `Clear all` / `Done` buttons, top-level `Apply`
- **New DOM that ships in `packages/core/src/components/control-bar/control-bar.ts:1391-1413`:**
  - `.query-builder-overlay` + `.query-builder-modal` (a `role="dialog"` titled "Filter Query Builder")
  - `<protspace-query-builder>` with AND/OR condition rows (see `query-condition-row.ts`, `query-value-picker.ts`)

`expect(filterMenu).toBeVisible()` at L470 times out at 5s because `.filter-menu` is never present.

Affected tests:
- `numeric header toggle reverses rendered bin order without changing color assignments`
- `linear numeric bins can realize fewer values than requested and filter stays aligned`
- `numeric filters reuse the shared unfiltered bin view when switching bins`
- `non-selected numeric annotations can still drive filtering`
- `numeric keyboard reordering persists and refreshes filter order immediately`
- `numeric tooltip shows the current display label after rebinning`
- `zero-match filters keep the numeric view empty instead of falling back to full data`
- `stale numeric filter labels are pruned when bins change`
- `loading a new dataset clears active filters before rendering the replacement data`

### Cluster B — `loadDemoDataset` (4 failures)

`loadDemoDataset` (L147-162) selects annotation `'ec'` after navigating to `/explore`. `selectAnnotation` then `waitForFunction` polls `selectedAnnotation === 'ec'` and times out at 60s. Most likely the default demo dataset no longer contains an `ec` column.

Affected tests:
- `categorical keyboard escape restores order and keeps non-manual sorting`
- `categorical keyboard reorder promotes to manual order and keeps Other fixed`
- `categorical pointer drag from size-asc keeps Other fixed and promotes to manual`
- `categorical pointer drag from alphabetical reverse promotes to manual order`

### Cluster C — keyboard escape poll (1 failure)

`keyboard escape after moving restores numeric order and value sorting` — the `expect.poll` at L1846-1847 never converges to the initial item order. May be a timing/flake or a real regression in numeric-keyboard escape handling. Needs trace inspection.

## Repro

```sh
pnpm dev:app                              # terminal 1
pnpm test:e2e --project=numeric-binning   # terminal 2
```

Final tally: **27 passed, 14 failed, ~12 min runtime.**

## Expected behavior

All 41 tests pass against `main`-equivalent UI. The filter and demo-dataset failures are test-debt from prior shipped refactors, not regressions in the components themselves.

## Effort estimate

- **Cluster A** (medium, ~2-3h): rewrite the six filter-menu helpers to drive the query builder. Each helper builds an equivalent query (e.g., `setFilterValues(annot, [a,b,c])` becomes "add a condition for annot, set operator, pick values, apply"). Test bodies that just call the helpers should keep working; tests asserting per-annotation toggle UI may need semantic updates.
- **Cluster B** (low, ~15-30m): swap `'ec'` for an annotation that exists in the current demo bundle (or load a stable fixture instead of `/explore`'s default).
- **Cluster C** (low-medium, ~30m): re-run with `--trace=on`, inspect the trace, fix or quarantine.

Total: ~3-4h.

## Notes

Discovered while reviewing #228 — the palette default change (`viridis` → `batlow`) and strategy default change (`linear` → `quantile`) flagged the test file as a test-debt hotspot. After fixing those default-driven assertions in #228, the 14 failures here remained and are independent.

tsenoner:
Fixes #237 — the `numeric-binning` Playwright tests broke after the filter-menu → query-builder rewrite. While in there, repaired the rest of the e2e suite (all pre-existing test-debt from shipped refactors) and made it run in parallel.

**What changed**
- `numeric-binning`: drive the query builder + isolation model instead of the removed `.filter-menu` DOM / `filteredProteinIds` state (#237).
- Repaired `isolation-dataset-swap` & `load-large-bundle` (columnar `_plotData` shape), `dataset-reload` (import-menu idempotency), `fasta-prep` (wired the orphaned project in + `MIN_SEQUENCES`/SSE/tour fixes), `figure-editor` (density-tolerant dot-size assertion), and `url-view-state` ×3 browsers (runtime demo-dataset discovery + encoding-robust URL / OPFS handling).
- Parallelized Playwright (`fullyParallel`, `workers: 50%`, `retries: 1`) and stabilized the two contention-sensitive tests.
- Upgraded `@playwright/test` 1.57 → 1.60.

**Result:** 148 passed, 1 skipped (WebKit has no OPFS), 0 flaky; full `pnpm test:e2e` ~37 min → ~3 min.

Partially addresses #249 (e2e slowness): the suite now runs in parallel (~37 → ~3 min). The "do we need all these tests / can we prune?" angle is out of scope here — all tests are kept (one added).

OpenSpec change for #237 lives in `openspec/changes/fix-numeric-binning-e2e/`.


jcoludar (approved):
## Review: ✅ Approve — sound test-debt repair

Reviewed via an adversarial multi-lens pass (false-green / weakened-assertion hunting, parallelism & flakiness, coverage vs #237 intent, dependency & config correctness, helper robustness), with each non-trivial finding independently verified against product source. **No blockers; every "major" was refuted or downgraded on verification.**

What held up well:
- The numeric-binning rewrite drives the real shipped query-builder + isolation API (`getCurrentData`/`isolateSelection`/`isolationMode`, zero-match disabled-Apply) via genuine clicks, not state pokes.
- The columnar `_plotData` adapters in `load-large-bundle` / `isolation-dataset-swap` faithfully mirror the canonical `materializePlotDataPoint` indexing.
- `@playwright/test` 1.57 → 1.60 is lockfile-consistent (root + app + `pnpm-lock`, integrity intact, no 1.57 leftovers); none of 1.60's removed APIs are used.
- Per-context OPFS/IndexedDB/localStorage partitioning defuses the most-feared shared-state collision under `fullyParallel`.

### Pushed as follow-up fixes (`c9f64af`) — names/spec/gated-config only; `tsc -b` (app + e2e) green
1. **`spec.md` zero-match scenario shipped a false assertion.** It claimed the view becomes *empty* after the query is applied, but the requirement directly above it — and the test — show `Apply & Isolate` is **disabled** at zero match, so nothing applies and the view stays unchanged. Rewrote the scenario to the real contract (0-of-N count, disabled Apply, view unchanged).
2. **Two test titles contradicted their assertions.** `"…shows the current display label after rebinning"` → `"…shows the raw value within the current bin range after rebinning"` (the test asserts the raw value, not a bin label — consistent with the test's own comment and `getProteinDisplayValues` returning the raw value when the bin-label map is empty). The zero-match title now states the disabled-Apply contract.
3. **`fasta-prep.live.spec.ts` was permanently dead.** The config comment promised a `fasta-prep-live` opt-in project that didn't exist. Wired it behind `RUN_LIVE_E2E` (excluded from the default suite, so the green run is unchanged).

### Non-blocking notes (tracked, not addressed in the pushed commit)
- **Real product bug — already being fixed in #259, so no action needed here.** `control-bar.ts` `_handleQueryApply` computes `matchedIndices` against full materialized data but maps them through possibly-isolated `getCurrentData().protein_ids`, so stacked applies map indices wrongly. The tests legitimately route around it via Reset-All (as `design.md` documents). #259 already refactors this exact path (drops the `matchedIndices.map(i => proteinIds[i])` mapping and adds `control-bar.query-apply.test.ts` with stacked-apply cases), so it's covered.
- **e2e assertion-quality follow-ups → tracked in #276.** Three minor, non-blocking items: the `figure-editor` dot-size floor (`* 1.2`) sits only ~7% above the measured ~1.29× (flake-prone + weak guard → prefer a monotonic 1×/2×/5× ladder); the renamed tooltip range check is global-min/max ± 2% (near-tautological → assert the value lands in its specific bin); and `url-view-state` runtime demo-discovery has no distinctness guard (could silently no-op if the demo ever collapses to one annotation/projection → add a fail-loud invariant in `beforeAll`). Details + suggested fixes in #276.

Note on the "0 flaky / retries:1" point that came up: CI's retry policy is unchanged from `main` (main already had `retries: 1` in CI); this PR only adds a *local* retry, and the two known flakes were root-caused in `197ce30`, not merely retried into green. Not a concern.

PR #280, opened 2026-06-13
fix: route Biocentral-unavailable failures to Colab

t03i:
## Why

Biocentral (the embedding backend) is routinely down. When it is, `protspace embed` exits with `TimeoutError: No healthy biocentral service became available in time` — a message the prep pipeline did **not** recognise. So the failure fell through to the generic *"The embedding step failed. Please try again…"* toast and users kept retrying a doomed job instead of being sent to the working Colab notebook.

Observed in a real job:
```
job failed … TimeoutError: No healthy biocentral service became available in time
reason='The embedding step failed. Please try again; …'
```

## What changed

- **Backend** (`pipeline.py`): add `"no healthy biocentral"` to `_BIOCENTRAL_DOWN_PATTERNS` so the outage timeout is classified `BIOCENTRAL_UNAVAILABLE`. Deliberately narrow — not bare `timeouterror` — so it can't swallow the legitimate pipeline wall-clock timeout.
- **Frontend notify primitive** (`notify.ts`): add optional `NotifyOptions.action` rendered as a Sonner action button (`mailto:` → `location.href`, `http(s)` → `window.open(_blank, noopener,noreferrer)`).
- **Frontend routing** (`notifications.ts`): rewrite the `BIOCENTRAL_UNAVAILABLE` copy to point to Colab and attach an **"Open in Colab ↗"** action for that code. Other failures are untouched.
- **De-dup** (`fasta-prep-limits.ts`): extract `COLAB_NOTEBOOK_URL`; `runtime.ts` now uses the shared constant.

## Coordination with `feat/contac_link`

The pending support-mailto design adds the same `NotifyOptions.action` field and a "Report this" action to the same builder. This PR implements that primitive **first**, in the exact shape that design specifies, and only owns the Biocentral arm of the failure toast (Colab wins the single action slot for a known, routine outage). Contact-link slots its "Report this" `else` arm on top with no conflict.

## Tests

- `test_pipeline.py`: new regression test for the "no healthy biocentral" timeout → `BIOCENTRAL_UNAVAILABLE`. Full prep suite green (80).
- `notify.test.ts` / `notifications.test.ts`: assert the action is forwarded and the Biocentral toast carries Colab copy + action. Frontend suite green (17).
- `type-check`, `knip`, `docs:build` pass (precommit).

Full design/spec/tasks under `openspec/changes/route-biocentral-down-to-colab/`.


tsenoner:
@peymanvahidi is this ready to merge into main?

peymanvahidi:
yes

PR #366, opened 2026-07-20
feat(bundle): contract-test the parquetbundle seam across Python and TypeScript

### LINKED ISSUE #334: Contract testing between frontend and backend
parquetbundle

t03i:
Closes #334.

> [!IMPORTANT]
> **After merging:** mark `Bundle format contract` as a required status check ([how](#after-merging)). Until that's done, a red contract run blocks nothing.

## Why

The `.parquetbundle` producer (`apps/protspace`, Python) and its consumer (`packages/core`, TypeScript) live in one repo, but nothing tested the path between them — and they had **already drifted**. `write_bundle` emits 3–5 parts while the reader rejected anything but 3–4, so `protspace bundle -s <statistics>` produced files the web app could not open at all.

Neither existing workflow could have caught it: `ci.yml` ignores `apps/protspace/**`, `protspace-ci.yml` only runs for it. A contract test in either is skipped by exactly the change most likely to break it.

## What changed

- **Reader** (`bundle.ts`) — accepts 3–5 parts, matching `_parse_bundle`. Statistics ignored; settings now bounded by the *next* delimiter rather than end-of-file, which previously glued the statistics part onto settings.
- **Writer** (`bundle-writer.ts`) — adds `assertNoBundleDelimiter`, mirroring Python's `_check_no_delimiter`, which had no TS counterpart. Without it, an export whose annotation text contains the delimiter loaded with all styling silently discarded.
- **Contract suite** (`tests/contract/`) — bundles generated at test time by the real `protspace bundle` CLI, so no fixture can go stale. All five producible layouts, plus a 12 000-row variant that reaches the optimized decode path.
- **CI** (`bundle-contract.yml`) — new workflow, unfiltered on every PR. A `paths:` filter cannot coexist with a required check: GitHub waits forever for a report a skipped workflow never sends.

## Verification

Contract 11 · utils 321 · core 1384 · web 165 · `pnpm quality` clean · green in CI.

Mutation-verified, not just observed green — removing either reader guard, removing the writer guard, or raising the optimized-path threshold above the fixture each turns the suite red.

## Notes for the reviewer

**One deferred item was withdrawn, not fixed.** A zero-byte settings part in a 4-part bundle no longer warns. Leave it: `bundle.py` branches on that part's *emptiness*, not on the part count, and the TS check matches it exactly. Conditioning on part count would make the reader branch on something the writer does not — a new drift, and one this suite could not catch. Detail in the archived change's `design.md`.

**Known gaps.** `tests/contract/` gets no static type-check (no `tsconfig` project reference). The suite asserts the fixture crosses the optimized-path threshold but can't verify that decoder actually ran without exporting module-private internals. `apps/prep/**` still has no CI coverage — same root cause as above, tracked separately.

## After merging

```bash
gh api --method PATCH repos/tsenoner/protspace/rulesets/12639946 --input - <<'JSON'
{ "rules": [
  { "type": "deletion" },
  { "type": "non_fast_forward" },
  { "type": "pull_request", "parameters": {
      "allowed_merge_methods": ["merge", "squash", "rebase"],
      "dismiss_stale_reviews_on_push": false, "require_code_owner_review": false,
      "require_last_push_approval": false, "required_approving_review_count": 0,
      "required_review_thread_resolution": false, "required_reviewers": [] } },
  { "type": "required_status_checks", "parameters": {
      "strict_required_status_checks_policy": false,
      "required_status_checks": [{ "context": "Bundle format contract" }] } }
] }
JSON
```

- Apply **after** merging — the unfiltered workflow must exist on `main` first, or every open PR blocks on a check that cannot report.
- `Bundle format contract` is the **job** name; GitHub matches that, not the workflow's `name:`.
- The live ruleset has a `bypass_actors` entry this payload omits. Re-read the ruleset afterward to confirm it survived.


tsenoner:
Reviewed this and pushed some changes — assigning back to you.

**One thing became a real fix rather than a deferral.** Deferred item #1 (the TS writer having no `_check_no_delimiter` counterpart) turned out to be worse than the original write-up said: widening the reader to 3–5 parts made one branch *silent*. A 4-part export contaminated inside the settings part used to be rejected outright, and would instead have loaded with all styling quietly discarded. Added `assertNoBundleDelimiter`, mutation-verified.

**Deferred item #2 I'd argue should stay unfixed**, and I've noted that in `design.md`. `bundle.py` branches on the settings part's *emptiness*, not the part count — so conditioning the TS skip on part count would make the reader branch on something the writer doesn't. That's a new drift, and one the contract suite couldn't catch. Happy to be argued out of it.

**Two CI defects found:** the test's own `uv run` was re-installing the dev group and undoing the workflow's `uv sync --no-dev` (CI logged `Removed 47247 files (5.9GiB)` on a job that reported installing 58 packages), and the job was silently running on CPython 3.14 because `python-version-file` pointed at `pyproject.toml`'s `>=3.10` rather than `.python-version`.

**The settings fixture was testing an unreachable branch** — `legendSettings` appears nowhere in the Python sources; the producer writes a flat `{annotation: envelope}` map. So the contract's only settings assertion exercised `isNormalizedBundleSettings` while the branch every real bundle takes went untested. Now flat.

Also removed the `paths:` filter entirely so the job can serve as a required status check — there's a post-merge step for that in the PR description.

t03i (commented):
Automated review of `main...HEAD`, cross-checked against the PR description and the follow-up comment. The two things you pre-empted are pre-empted correctly — I verified both rather than taking them on trust — so this is only about what's left.

**Answered by the thread, no action:**

- *Zero-byte settings part no longer warns.* Your rationale holds against the source: `_parse_bundle` (`bundle.py:53`) is `parts[3] if len(parts) >= 4 and parts[3] else None` — emptiness, not part count. Conditioning the TS skip on `delimiterPositions.length === 4` would be a genuine new drift. Withdrawal is right. (One residual worth a line in `design.md`: the property both sides now share is that a *truncated* 4-part bundle loads with default styling and no diagnostic. Symmetric, so not a regression here — but it is a format-level hole, not just a TS one.)
- *`tests/contract/` gets no static type-check.* Already in Known gaps. It is also outside `eslint.config.mjs`'s `files` globs, which the writeup doesn't mention — worth adding to the same gap note so it's one item, not two discoveries.
- *In-band delimiter inside a 4-part settings part.* You identified the exact mechanism and shipped the write-side guard. Residual only: bundles already exported by the shipped app can carry the delimiter, and for those the widened reader turns a hard `found 4` rejection into a silent settings drop (`extractSettings` fails magic → swallowed by its `catch` → `console.warn`). The guard closes new files, not files already on disk. Probably acceptable; flagging so it's a decision rather than an oversight.

**Not addressed anywhere — see inline comments:**

1. Statistics part is read-but-not-carried, and the TS writer can't emit it → load/edit/export in the web app silently drops it (the Python `replace_*_in_bundle` helpers carry it byte-for-byte). This is the one I'd want a call on before merge — the PR makes previously-unopenable files openable, and openable now means lossy.
2. `spawnSync` with no `timeout` — `hookTimeout: 300_000` can't fire while the main thread is blocked in a synchronous child, and the job has no `timeout-minutes`. On an unfiltered required check, a stalled `uv` pins a runner to the 6-hour default and blocks every open PR.
3. `assertNoBundleDelimiter` → `isParquetBundle` → `findBundleDelimiterPositions` collects *every* match with no early exit, run synchronously over every part. On the SwissProt-scale export the reader's own comment cites, that's a second full pass over hundreds of MB for a boolean.
4. The guard's throw names neither the offending feature nor the offending value, and `export-handler.ts` surfaces it generically — a user with a contaminated annotation label has no way to find it and no way to export at all.
5. Fixture constants hand-mirrored between `emit_bundles.py` and the TS consumer.

**Considered and dropped** (recording so they don't get re-raised): missing `concurrency` group — `ci.yml` doesn't have one either, so it's repo convention, not a deviation in this PR. `setup-python` layered on `setup-uv` — checked against uv 0.11.28 behaviour; uv does fall through to the search path, so the pin works as documented. No in-process JS test for the 5-part accept path — mostly covered by the contract suite running unfiltered on every PR, though `bundle-roundtrip.test.ts` already builds real parquet parts, so the stated obstacle ("with mock parts the call still rejects during decode") doesn't actually apply if you ever want the fast local signal.

packages/core/src/components/data-loader/utils/bundle.ts
```diff
@@ -56,46 +56,51 @@ function readFormatVersion(metadata: FileMetaData): number {
...
 export async function extractRowsFromParquetBundle(
   arrayBuffer: ArrayBuffer,
 ): Promise<BundleExtractionResult> {
   const uint8Array = new Uint8Array(arrayBuffer);
   const delimiterPositions = findBundleDelimiterPositions(uint8Array);
 
-  // Support both 2 delimiters (original) and 3 delimiters (with settings)
-  if (delimiterPositions.length !== 2 && delimiterPositions.length !== 3) {
+  // 2 delimiters (original), 3 (with settings), or 4 (settings + statistics).
+  if (delimiterPositions.length < 2 || delimiterPositions.length > 4) {
```
t03i, line 78: **Statistics survive the read but not a round-trip.**

Widening to 4 delimiters makes 5-part bundles loadable, but nothing consumes part 5 and `createParquetBundle` (`bundle-writer.ts:296-301`) can only ever push 3 or 4 buffers. So: user runs `protspace bundle ... -s statistics.parquet`, opens the file in the web app (which before this PR simply refused it), recolors a legend, hits Export — and the trustworthiness/continuity metrics are gone from their only copy, silently. The Python `replace_settings_in_bundle` / `replace_annotations_in_bundle` deliberately carry that part through byte-for-byte, so the web app becomes the only lossy hop in the format.

Cheapest fix that closes it: return the raw `part5` buffer from `extractRowsFromParquetBundle` alongside `settings`, thread it into `VisualizationData`, and have `createParquetBundle` re-append it when present. If that's more than this PR wants to carry, the alternative is to warn at export time when the loaded bundle had a statistics part — better than dropping it quietly.

packages/utils/src/parquet/bundle-writer.ts
```diff
@@ -295,6 +296,12 @@ export function createParquetBundle(
     buffers.push(settingsBuffer);
   }
 
+  // The delimiter is in-band and unescaped, so a part containing it would split
+  // into two on read-back. The Python producer guards every part it writes; do
+  // the same here or the invariant holds in only one direction. Annotation text
+  // and legend category names are user-authored, so this is reachable.
+  buffers.forEach(assertNoBundleDelimiter);
```
t03i: The guard is right to exist, but as written a contaminated dataset means the user cannot export **at all**, and the message names neither the offending annotation feature nor the offending value — `export-handler.ts` just surfaces it as a generic failure. Annotation text is user-authored and imported from arbitrary CSV/metadata, so this is reachable, and from the UI there's no way to locate the bad label.

Since the parts are built from `data` right here, the check could run per-part and say which one failed (`annotations` / `projections metadata` / `projections` / `settings`) at near-zero cost:

```ts
const parts = { annotations: annotationsBuffer, ... };
for (const [name, buf] of Object.entries(parts)) {
  if (isParquetBundle(buf)) throw new Error(`The ${name} part contains the reserved delimiter ...`);
}
```

Still not the offending *value*, but it turns "export is broken" into "your annotations are the problem".

packages/utils/src/parquet/delimiter-utils.ts
```diff
@@ -29,6 +29,29 @@ export function findBundleDelimiterPositions(uint8Array: Uint8Array): number[] {
...
+ *
+ * Mirrors `_check_no_delimiter` in the Python producer
+ * (`apps/protspace/src/protspace/data/io/bundle.py`) — both sides must enforce
+ * this or the format's invariants hold only in one direction.
+ *
+ * @param arrayBuffer - The serialized part to check
+ * @throws If the part contains the reserved delimiter byte string
+ */
+export function assertNoBundleDelimiter(arrayBuffer: ArrayBuffer): void {
+  if (isParquetBundle(arrayBuffer)) {
```
t03i, line 53: `isParquetBundle` is `findBundleDelimiterPositions(...).length > 0` — the naive O(N·M) scan walks the entire part and materializes an array of *every* match, when this caller only needs "is there at least one". Invoked synchronously on the UI thread for all 3-4 parts, right after `createParquetBundle` has already built hundreds of MB (the reader's own comment cites SwissProt 573 K / ~2.3 GB peak), it's a second full pass that extends the frozen-tab window during export.

A `containsBundleDelimiter` that returns on the first match — or just reusing the existing scan with an early-exit flag — keeps the guard and drops the worst case to first-occurrence.

tests/contract/bundle.contract.test.ts
```diff
@@ -0,0 +1,258 @@
...
+        '--package',
+        'protspace',
+        '--no-dev',
+        '--locked',
+        'python',
+        'tests/contract/emit_bundles.py',
+        outDir,
+      ],
+      { cwd: REPO_ROOT, encoding: 'utf-8' },
+    );
```
t03i, line 85: No `timeout` on the child, and vitest's `hookTimeout: 300_000` cannot rescue it — a timer can't run while the main thread is blocked inside `spawnSync`. The workflow sets no `timeout-minutes` either, so a stalled `uv` (index outage, cache hiccup — and `uv run` re-resolves on every invocation) runs to GitHub's 6-hour default. On a check that is deliberately unfiltered and about to be marked required, that means every open PR sits un-mergeable for six hours.

```ts
{ cwd: REPO_ROOT, encoding: 'utf-8', timeout: 240_000 }
```

plus `timeout-minutes: 10` on the job. Worth pairing with `maxBuffer` — the default is 1 MiB and a chatty producer traceback would be truncated in exactly the failure you most need to read.

tests/contract/bundle.contract.test.ts
```diff
@@ -0,0 +1,258 @@
...
+// Imported by source path, like the core reader above: the suite tests the
+// working tree, not built package output. (The vitest config still aliases
+// `@protspace/utils` — packages/core's own sources import it that way.)
+import { BUNDLE_DELIMITER_BYTES } from '../../packages/utils/src/parquet/constants';
+
+const REPO_ROOT = resolve(__dirname, '../..');
+const PROTEIN_COUNT = 10;
+const PROJECTION_COUNT = 2;
+/** Mirrors LARGE_PROTEIN_COUNT in emit_bundles.py. */
+const LARGE_PROTEIN_COUNT = 6_000;
```
t03i: `PROTEIN_COUNT`, `LARGE_PROTEIN_COUNT`, `LABEL_WITH_RESERVED_CHAR` and the implicit null-at-index-3 convention are hand-mirrored from `emit_bundles.py`, with `/** Mirrors ... */` comments doing the enforcing. Bump `LARGE_PROTEIN_COUNT` on the Python side because the optimized-path threshold moved, and the suite fails as `expected 6000, received 8000` in the *consumer* — pointing at the reader rather than at the generator that actually changed.

Having `emit_bundles.py` drop a small `manifest.json` next to the bundles (counts, the label, the null row index) and reading it in `beforeAll` deletes all four constants and keeps the generator as the single source of truth — the same reasoning already applied to `OPTIMIZED_PATH_ROW_THRESHOLD`.

PR #381, opened 2026-07-23
feat(protspace): require Python >=3.12, fix the CI version matrix, add a future-Python canary

tsenoner:
## Why

Investigating "why is protspace only compatible with 3.10–3.12" turned up that the premise was wrong in **both** directions:

- **Nothing blocks 3.13 or 3.14.** Real 3.13.14 and 3.14.6 venvs run the full fast suite identically (787 passed), including numba-JIT UMAP/PaCMAP. Every core dep ships cp313+cp314 wheels or is pure-py3. The historical blocker (umap-learn → numba/llvmlite) is gone.
- **The 3.10/3.11 support we advertised was already broken.** The shared root `uv.lock` has been `>=3.12` for a while because `apps/prep` (a workspace member) forces that floor, so `uv sync --python 3.10` from `apps/protspace` hard-errors. Those versions were unbuildable through the workspace and never actually tested.

And the CI matrix that claimed to cover 3.10/3.11/3.12 **ran 3.12 three times**: `actions/setup-python` installed the matrix interpreter, then a bare `uv sync` discarded it (below the `>=3.12` lock floor) and silently fell back to system Python 3.12. All three legs reported an identical green run. Proof from the last main CI run:

```
test (3.10) | Successfully set up CPython (3.10.20)
test (3.10) | Using CPython 3.12.3 interpreter at: /usr/bin/python3   ← not 3.10
test (3.11) | Using CPython 3.12.3 interpreter at: /usr/bin/python3   ← not 3.11
test (3.12) | Using CPython 3.12.13 …                                 ← correct
```

`lint`/`publish`/`release` had the mirror-image bug: `python-version-file` resolves `>=3.12` to the *newest* available, so lint was running on **3.14.6**, drifting with the runner image.

## What this changes

**`feat`: require Python `>=3.12`** — `protspace` + `protlabel` declared floor now matches what is actually built and tested. `uv.lock` is unchanged (prep already pinned the intersection). Versioned as a **minor** bump (4.8.2 → 4.9.0) per maintainer decision.

**`ci`: make the matrix honest and future-aware**
- Drive the interpreter through `setup-uv`'s `python-version` (sets `UV_PYTHON`) so the matrix axis actually binds; drop `actions/setup-python`.
- Retarget the matrix to the real supported range: **3.12, 3.13, 3.14** (`fail-fast: false`).
- Add an **assert step** that fails loudly if the running interpreter ever differs from the matrix version again — the guard that would have caught the original bug.
- `uv sync --locked` so a stale lock fails instead of silently re-resolving.
- Pin lint/publish/release to 3.12 the same way (no more floating to the newest runner Python).
- Move the inert `apps/protspace/.python-version` → workspace root, where uv actually reads it.
- **New `protspace-future-python.yml`**: a monthly + on-demand canary that fresh-resolves the dev env on the **next** Python (3.15) and fails loudly, so a coming break surfaces before a user hits it.

**`refactor`: py312 idioms** the stricter ruff target unmasked — the three Typer CLI enums → `enum.StrEnum` (behaviour-preserving: name == value, consumers use `.value`/`==`), and `datetime.timezone.utc` → `datetime.UTC`.

## Are we future-proof now?

Partly, and honestly so. The matrix tests everything **currently released** (3.12–3.14). The canary gives **early warning** for the next release (3.15) — GitHub emails on scheduled-workflow failure. It is expected to be *red until the scientific stack ships 3.15 wheels* (numba/llvmlite are always last), so an early red there usually means "not ready yet", not a protspace bug — read the log. Adding a new version to the matrix stays a deliberate one-line change once the canary is green; nothing auto-adds versions, by design.

## Verification

- `uv sync --locked --python {3.12,3.13,3.14}` → clean (exit 0) against the current lock.
- Fast suite: **787 passed, 6 deselected** on 3.12 (and independently reproduced on 3.13.14 / 3.14.6 during the audit).
- `ruff check` + `ruff format --check` clean on py312 target; `pnpm precommit` green.
- All 9 workflow YAMLs parse.

## Follow-ups (not in this PR)

- **Two dead-code defects** found en route, unrelated to versions: `reducers.py` `_ensure_annoy_or_fallback()` is dead (pacmap 0.9.1 swapped annoy → faiss-cpu) and logs a spurious "annoy is non-functional" warning on every PaCMAP run; `scripts/generate_examples/generate.py` has a `tomli` fallback for a package that's declared nowhere. Will send as a separate PR.
- The `[local]` extra (torch/transformers) has no CI coverage on any Python — the e2e tests are `slow` + `importorskip`.


tsenoner:
Superseded by #382. This PR auto-closed when its branch was renamed off the `t03i/` namespace (a colleague's) to `feat/python-version-support`. Same commits, same green CI — continue in #382.

Issue #379, opened 2026-07-22
[BUG] Jupyter Notebook release link broken


t03i:
### Bug Description

Dataset release link in the notebooks still assumes the old repository. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

tsenoner:
Fixed by republishing the `examples` release onto the monorepo, so the notebook URLs work unchanged.

All 10 assets mirrored byte-identical from `protspace-legacy` (SHA-256 verified) and confirmed serving 200. Marked `--latest=false`, so `v4.9.1` remains the Latest release.

Issue #320, opened 2026-06-17
[FEATURE] The Google colab notebook should be independent of Biocentral


tsenoner:
Update the google colab notebook to use its GPU instead of the Biocentral in order to always work when it is down.

tsenoner:
Implemented and merged as of 2026-07-17 via the local-embedding-backend / Biocentral-independence work:

- #332 — local GPU/CPU embedding backend + biocentral/local switch
- #349 — local↔Biocentral embedding parity verification
- #350 — local embedding backend in the Preparation Colab
- #351 — projection statistics toggle wired into the Colab
- #352 — optional in-session EAT cell in the Colab

The Colab notebook can now embed locally on its own GPU/CPU and no longer depends on Biocentral being up. (Note: these PRs referenced "issue #59" in their titles by mistake — #320 is the correct tracking issue.) Closing as completed.

Issue #318, opened 2026-07-02
[FEATURE] Optimal-k selection: gap statistic + BIC/AIC (GMM)


tsenoner:
Follow-up split out from the projection-statistics work (parent: tsenoner/protspace#324, which lists four optimal-*k* methods).

## Status
`--cluster-selection` currently supports **`elbow`** (inertia knee) and **`silhouette`** (max mean silhouette). The remaining two from tsenoner/protspace#324 are **deferred**:

- **Gap statistic** — compare within-cluster dispersion to *B* uniform null-reference clusterings; pick the smallest *k* whose gap exceeds the next *k*'s (minus its s.d.).
- **BIC / AIC via GMM** — fit a Gaussian Mixture per *k*, pick the information-criterion minimum.

## Why deferred
- The auto-clustering is the **secondary** deliverable now that cluster-validity is annotation-based (see the annotation-cluster-validity rework); adding more K-selection methods to it is scope creep for that PR.
- **Gap** needs *B* reference clusterings per *k* → costly, awkward at 570k scale (would need aggressive subsampling to stay bounded).
- **BIC/AIC** requires a separate **GMM** code path (a different model from the KMeans used everywhere else).

## Integration (when picked up)
Both slot in additively as new `--cluster-selection` values (`gap`, `bic`), producing a `cluster_<method>_<proj>` membership column + a `kmeans_<method>` / `gmm_<method>` `label_kind` in `statistics.parquet` — no rework of the existing elbow/silhouette paths. Reuse the existing subsample / hard-ceiling guards for scale.

BIC-via-GMM is the more principled of the two if only one is added.

Refs: tsenoner/protspace#324.

Issue #74, opened 2025-10-27
[FEATURE] Retrieve additional information when hovering over a dot in the canvas


tsenoner:
### Is your feature request related to a problem?

I would like to get additional information about the likelihood of InterPro predictions.

### Describe the solution you'd like

When hovering over a dot in the canvas and an InterPro feature is selected, display the prediction score next to the prediction. For instance, for Pfam predictions.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #391, opened 2026-07-27
fix(notebook): unbreak both Colab notebooks, move MMseqs2 to an extra, fix stale refs

### LINKED ISSUE #392: [BUG] ProtSpace_Transfer notebook has no data-download step
### Bug Description

`apps/protspace/notebooks/ProtSpace_Transfer.ipynb` cannot run top-to-bottom in Colab. It has no data-download step, so the very first command fails on a fresh runtime.

The cells are: install → `protspace transfer` → read results. The transfer cell references two files that are never created:

```bash
!protspace transfer \
  -b results.parquetbundle \
  -e embeddings.h5:prot_t5 \
  ...
```

Unlike `ProtSpace_Preparation.ipynb`, which downloads an example `.h5` from the `examples` release, this notebook assumes `results.parquetbundle` and `embeddings.h5` are already present.

It also transfers `protein_category` filtered on `--query-id-prefix TRINITY_`, which looks carried over from a specific local dataset rather than anything a reader can obtain.

### Suggested fix

Add a download cell pulling `snake_toxin.parquetbundle` + `snake_toxin_prot_t5.h5` from the `examples` release (both now live at `tsenoner/protspace`), and retarget the transfer to an annotation/filter that actually exists in that bundle. The bundle's schema should be confirmed first — I have not verified which columns it carries.

Found while auditing #379; unrelated to that fix.

tsenoner:
Cleanup after the monorepo rename, plus the Colab notebook problems found while verifying it.

Closes #392. Related to #379 (fixed separately by republishing the `examples` release).

> **Merged `main` mid-flight.** A docs consolidation landed on `main` while this was open and turned the two reference docs this branch was editing — `apps/protspace/docs/cli.md` and `docs/annotations.md` — into stubs pointing at protspace.app. The merge (3596f99e) takes those stubs and re-targets this branch's work onto the surviving copies under `docs/guide/`. Two of this branch's edits became moot and were dropped: the stale-ref fix in `annotations.md` (`main` deleted the CATH passage it corrected) and the ESM-C licensing fix in `cli.md` (which landed on `main` separately as dafe7208). So parts of the story below are now `main`'s work rather than this PR's — noted where that is the case.

> **History rewritten** (early on the branch). A 2.4 MB `.h5` was committed to the repo root by mistake. Since this PR is not squash-merged, deleting it in a later commit would have left the blob in `main` forever, so the branch history was rewritten to purge it and force-pushed. The tree was byte-identical across the rewrite; only the SHAs changed.

## ⚠️ Upgrade note

`-s/--similarity` now requires `pip install "protspace[similarity]"`.

MMseqs2 (`pymmseqs`) moved out of the base install. It was reachable only through that one flag, yet every release through 1.1.0 shipped **cp310-only wheels** — so against this package's `requires-python = ">=3.12"` it compiled from source for every user on every platform. It also pinned `ipython<9`, which is what upgraded Colab's pinned ipython and produced the dependency-conflict wall in the notebook.

> Worth knowing: `pymmseqs` 1.2.0 was published on 2026-08-11, after this work, and finally ships `py3-none-*` wheels and drops the ipython pin. `uv.lock` still resolves 1.1.0, so nothing here changes today — but the wheel argument for keeping it out of the base install has an expiry date now, and the extra can be revisited once the lock moves.

The check runs as an argument precondition in both `prepare` and `project`, so it fails **before** any embedding work rather than at the `pymmseqs` import site — otherwise `prepare -i seqs.fasta -e prot_t5 -s` would pay a full embed run before being told to `pip install`. It raises `typer.BadParameter`, so the hint arrives as a clean `Error:` line instead of a traceback. `prepare`'s *other* `-s` precondition ("requires FASTA — use `-f` when input is HDF5") moved up alongside it, so both pure-argument checks now run before any input is read.

`similarity.py` keeps an `ImportError` backstop for direct library callers, but only for `pymmseqs`' own absence: `pymmseqs` pulls in ipython/pandas/scikit-learn, and translating *every* `ModuleNotFoundError` into "MMseqs2 is not installed" told a user with a broken-but-present install to install what they already had — while `find_spec("pymmseqs")` succeeded, so the fast-fail guard waved them through and they paid a full embed run first. Anything that is not `pymmseqs` now re-raises with the real module name.

Shipping as a 4.x release by decision, so there is deliberately no `BREAKING CHANGE:` footer. The note lives in the package README and the published CLI guide, since `CHANGELOG.md` is regenerated wholesale by semantic-release.

Also relaxed two of our own floors that caused **5 of 6** conflict warnings: `rich>=13.8.0` (2 lines, matching what typer already requires) and `protobuf>=3.20` (3 lines). Neither is imported anywhere. protobuf keeps a deliberately low floor rather than no specifier at all — `>=3.20` is the real lower bound for transformers' slow-tokenizer conversion (ProtT5/Ankh `T5Tokenizer`), and every other entry in the `local` extra carries one. `dash-treeview-antd` was dropped from the dev group as unused.

## `--help` was hiding the extra it tells you to install

Rich parses a bracketed word as a markup tag and silently drops it, so `--backend` had been rendering ``needs `pip install protspace` `` — with the extra missing entirely — for as long as that option has existed. That string and the `[similarity]` mention added here are both escaped now, so each flag that requires an extra actually names it.

Inside the CLI the embedder model list also collapsed from three copies to one. It existed as a validation set in `prepare.py` plus verbatim help strings in `prepare.py` and `embed.py`, which is why correcting a single licence sentence took two identical edits in this very PR. It stays a literal rather than deriving from `ALL_SHORT_KEYS`, because importing `data.embedding.biocentral` drags h5py and biocentral_api into every CLI startup including `--help` — about 180 ms on top of what the CLI already loads. A test fails if the two drift.

## Both Colab notebooks' setup cells never ran

`%%capture` sat below `# @title`, and a cell magic is only recognised on line 1. IPython parses it as the line magic `%capture`, which doesn't exist, so the cell raised `UsageError` and aborted before `pip install` and every import — taking all later cells down with `NameError`. Reproduced on IPython 7.34.0 (Colab's pin) and 8.39.0; present since each notebook was added.

Installs now go through `subprocess.run(..., capture_output=True)`: position-independent, keeps `# @title` on line 1 where Colab needs it, and immune to whatever pip config Colab has. Both streams are captured and dropped on success, so a clean install prints nothing; on a non-zero exit the last 3000 characters of each are raised as the `SystemExit` message. The rationale comment is byte-identical in all three notebooks — it had drifted into three versions (five lines, three lines, and none), and the copy with none is the one most likely to be "simplified" back into the broken form.

The prep notebook's restart hint now wraps the whole import block. It used to start below `import h5py` / `pandas` / `tqdm` — the preloaded compiled extensions the install is most likely to break, and exactly the case the hint exists for — so that case produced a raw traceback instead of "Runtime > Restart session".

## Notebook UX

- **One control panel.** Config was split across two widget cells, which read as "configure → run → configure → run" even though the handlers already read widget values at click time. Merged into a single `Accordion` with Generate always visible outside it, so the panel can't outgrow Colab's output cap and trap the input tabs behind a nested scrollbar.
- **"Download" meant two opposite directions** — fetching example data *into* the runtime and saving the bundle *out* to your computer. One verb per direction now, stated once in the header and applied throughout.
- **The licence sentence is imported, not retyped.** The panel had its own copy, which would have made a fourth copy of a string `CLAUDE.md` already tracks as "keep the three in step" — in the PR that exists partly because correcting it took two identical edits. It now imports `EMBEDDER_HELP_LICENSE` and `EMBEDDER_MODELS`.
- **Transfer notebook had no data step** (#392). It ran against `results.parquetbundle` and `embeddings.h5` that nothing created. Now fetches the Snake Toxin example and transfers `protein_families` from Swiss-Prot to TrEMBL — verified end to end on the real release assets: 838 rows filled at ~0.97 confidence.
- **Interrupted downloads no longer poison the runtime.** `urlretrieve` streams into its destination, so an aborted transfer leaves a partial file that satisfies `.exists()` and is reused forever, killing step 3 with `OSError: truncated file` while the cell prints "Reusing embeddings.h5". Downloads land on a `.part` file renamed only once complete. Both files are cached under their own names and the 320 KB bundle is copied to the working path each run, so step 3 still starts from a pristine copy *and* a warm re-run needs no network. Verified cold, warm, interrupted and post-step-3 re-run against the real assets: 22,402,488 bytes restored after truncation, re-run bundle byte-identical.
- **The MDS note shows whenever MDS is selected.** It was keyed to "no parameter group is visible", so picking MDS alongside UMAP — when the note is most worth reading — hid it.
- Merge residue removed from the control panel: a leftover `# @markdown` header telling users to "Run this cell, then pick a tab" three lines under the new one saying they never need to; a `sim_cb` checkbox that was rendered but never read (and whose feature now needs an extra the notebook doesn't install); a re-hardcoded ESM-C pair that `_ESMC` already held; and a mid-cell `from ipywidgets import (...)` that switched namespaces halfway, shadowing the `IHTML` alias that exists precisely to avoid that collision.

## Stale `#N` references

Bare refs written in the standalone repo now resolve against the inherited `protspace_web` numbering, landing on unrelated frontend items — each returning HTTP 200, so nothing signals the wrong target.

Checking every number against the live API turned up something a first pass missed: **three issues were _transferred_ into the monorepo** rather than merely stranded, so they have new numbers here and the old ones are simply wrong.

| old ref | now |
| --- | --- |
| `#31` | `#324` (transferred) |
| `#59` | `#320` (transferred) |
| `#64` | `#318` (transferred) |
| `#10` `#44` `#54` `#55` `#56` `#57` `#58` `#60` `#63` `#66` `#67` `#68` `#73` `#74` | `tsenoner/protspace-legacy#N` |

About 50 references across source, tests, a notebook and thirteen design docs, including nine hyperlinks that pointed at `tsenoner/protspace/issues|pull/<N>` — worse than a bare ref, because they render as authoritative links to the wrong repo. Five `protspace_web#N` refs were also normalised to bare `#N`: those numbers *are* this repo's, so the old repo name in front of them was the stale part.

Docs are qualified at each file's definitional line (`**Issues:**`, `**Refs:**`, `**Trigger:**`) rather than in every sentence; fully qualifying prose turns headings into `### 2.3 tsenoner/protspace-legacy#57: ...`. Two docs that carried bare refs with no definitional line at all got one. Deliberately untouched: `#1`–`#6` in the EAT-UX docs are list markers, not issue refs; and `CHANGELOG.md` is regenerated by semantic-release, so edits there wouldn't survive.

`AGENTS.md` now records the convention, so the next doc doesn't reintroduce the problem: what bare `#N` resolves to, the three transfer mappings, and the qualify-once-per-doc rule — with code comments and docstrings always qualifying in full, since no definitional line travels with them.

## ESM-C is MIT now

Relicensed 2026-05-27 when it moved to Chan Zuckerberg Biohub, retroactively covering the Dec-2024 checkpoints; Synthyra passed the grant through. `esmc_600m` had not been non-commercial for two months while we still said it was.

`main` has since landed the same correction to the `embed`/`prepare` help and `docs/guide/python-cli.md`, so what remains here is the agent doc: `apps/protspace/CLAUDE.md` gets the merged note — `main`'s wording plus this branch's rationale for staying on the Synthyra checkpoints, which `main` had nowhere. Not switching to the official `biohub/ESMC-*` weights: transformers has no `esmc` type (port #46419 still open) and those repos ship no remote code, so it needs the `esm` SDK — pinned to `transformers<4.48.2` and Python `<3.13`, with git-URL deps PyPI forbids in published distributions.

## Tests, because none of this was catchable

Three new test files, ~296 lines. The point is that every fix above was previously guarded by nothing but a comment.

- **`test_notebooks.py`** globs `notebooks/` rather than naming files: no cell magic below line 1, every code cell compiles after IPython transformation, cell ids present while `nbformat_minor >= 5`. This is the only thing standing between the repo and a repeat of the `%%capture` bug — `.prettierignore` excludes `**/*.ipynb`, lint-staged has no notebook glob, and CI runs ruff over `src/ packages/ tests/` only. Verified by reinserting the real bug.
- **`test_docs_extras_sync.py`** pins the extras section identical between `apps/protspace/README.md` (PyPI) and `docs/guide/python-cli.md` (protspace.app), which nothing else compares, and pins the guide's embedder shortcut list to `EMBEDDER_MODELS` rather than to another copy of itself.
- **`test_cli_no_similarity.py`** covers the `-s` guard end to end: it fires before any load or embed, the loader backstop, and `EMBEDDER_MODELS` matching the embedder registry.

`docs/guide/python-cli.md` joins `protspace-ci.yml`'s path filter, because that is the only workflow that runs pytest — a PR editing just the docs-site copy would otherwise have skipped the pin protecting it.

## Verification

`pnpm precommit` green (it runs on every commit via husky), **862 Python tests pass** (`uv run pytest -q -m "not slow"` from `apps/protspace`, covering both workspace members; `pytest tests/` alone reports 825). ruff check + format clean, prettier clean, type-check and lint clean.

The three CLI guard tests initially passed locally and failed on all of 3.12/3.13/3.14 in CI: GitHub Actions sets `FORCE_COLOR`, so Rich emits colour *inside* the rendered message and the hint stops being a contiguous substring. Reproduced locally with `FORCE_COLOR=1`, fixed by stripping ANSI alongside the existing box-glyph and wrapping normalisation, and re-checked at `COLUMNS=60`.

`.gitignore` additions are `*.h5`, `*.part`, `/*.parquetbundle` and `/output/`, checked to shadow no tracked file. 20 `.parquetbundle` fixtures stay visible, none of them at the repo root — which is what makes that rule safe to anchor there while `*.h5` is repo-wide. `*.part` is the download staging name introduced here, so an interrupted fetch is exactly when one is left behind.

https://claude.ai/code/session_01Ar6ME3o6VYLYsie3J4kxE5


peymanvahidi (approved):

peymanvahidi:
Looks good to me

PR #389, opened 2026-07-26
Show annotation quality stats in the color-by dropdown

peymanvahidi:
The backend now ships optional projection-quality stats in the `.parquetbundle`.
In the frontend: every annotation in the color-by dropdown with at least one stat gets an ⓘ.
You can open it for:

- **Separation in <projection>**, how cleanly the annotation's categories sit in the current projection, each metric against the source-embedding ceiling (`emb 0.34`) so you can see what the projection kept vs. lost. Lower-is-better metrics get a ↓.
- **Auto-cluster agreement**, ARI/NMI between the auto-clusters and the annotation.
- A note that scores are on the full dataset (they don't recompute on isolate/filter/hide, so we say so).

Annotations without stats, and bundles built without `--stats`, look exactly as before, all opt-in and drawn defensively. Under the hood: read 4-/5-part bundles instead of rejecting them, round-trip the stats part on export, and a new `annotationStatSummary` selector in `@protspace/utils`.

fixes #296

fixes #330

PR #394, opened 2026-07-27
fix(bundle): close the remaining lossy paths in the frontend .parquetbundle export

### LINKED ISSUE #303: [BUG] Frontend .parquetbundle export is lossy — drops scores/evidence/secondary hits (first-label only)
## Context

When the web app exports the current view to a `.parquetbundle`, the annotations table is built by `createAnnotationsParquet` in `packages/utils/src/parquet/bundle-writer.ts`. For each categorical annotation column it writes **only the first hit's display label**:

- `getFirstAnnotationIndex(annotationIndices, i)` → `annotation.values[idx]` (`bundle-writer.ts:59-62`).
- All **secondary hits**, all **scores** (`annotation_scores`), and all **evidence codes** (`annotation_evidence`) are dropped.
- The label is written **decoded** and is **not re-serialized** into the `accession (name)|score;...` grammar.

So a protein whose `pfam` cell was `PF00001 (7tm_1)|50.2;PF00002 (7tm_2)|60.5` re-exports as just `PF00001 (7tm_1)` — silent data loss on round-trip. (`bundle-roundtrip.test.ts` already documents this: exported `annotation_scores`/`annotation_evidence` come back `{}`.)

## Why it matters

- Users who export → re-import lose multi-hit domains, bit scores / E-values, and evidence codes.
- It compounds after the bundle **format v2** encoding redesign (tsenoner/protspace#56 / #57 / #58): the backend will produce lossless, percent-encoded, versioned cells, but the frontend re-exporter would still flatten them to a single decoded label — and since it writes labels **without re-encoding**, an exported label containing a `;` is not even v2-safe.

## Proposed fix (natural follow-up once v2 exists)

- Re-serialize each categorical cell from the decoded `{labels, scores, evidence}` structures back into the v2 grammar via the shared annotation codec (encode names, join hits with `;`, emit `|score` / `|EVIDENCE`), preserving all hits + scores + evidence.
- Stamp exported bundles with `format_version = 2` in parquet key-value metadata, matching the backend writer.
- Extend `bundle-roundtrip.test.ts` to assert scores / evidence / secondary hits survive export → import.

## Related

- Encoding redesign spec: `tsenoner/protspace` → `docs/superpowers/specs/2026-07-08-annotation-encoding-redesign-design.md` (§7 explicitly scopes this out of the core fix).
- tsenoner/protspace#56, #57, #58.

## Acceptance criteria

- Frontend export → import round-trips categorical annotations losslessly (all hits, scores, evidence).
- Exported bundles carry `format_version = 2` and decode correctly on re-import.

tsenoner:
Closes #303.

## What #303 reported vs. what was left

The headline of #303 — first-label-only export, dropping secondary hits, scores, evidence, and `format_version` — was **already fixed** incidentally by the EAT / format-v2 work (`5b1f9fda`…`d53bc50e`), and is covered by `v2-roundtrip.test.ts`. The issue's cited evidence ("exported `annotation_scores`/`annotation_evidence` come back `{}`") is stale.

An audit of the remaining export → import path turned up four real losses, fixed here.

## 1. `__NA__` sentinel written instead of NULL (`156e3bdd`)

A missing cell is materialised on import as the in-memory `__NA__` category (`appendSyntheticNACategory`). The writer only skipped `value == null`, so the sentinel was serialised as a literal 6-character string. `__NA__` is not a `MISSING_VALUE_TOKEN`, so it re-imported as a genuine frequency-sorted category, took a palette slot away from `NA_DEFAULT_COLOR`, and was scored as a real class by `protspace stats`.

Now dropped when rebuilding the wire cell, mirroring the read side's `readCategoricalStorageValues`.

## 2. Python could not read the frontend's settings part (`1c33e62e`)

The frontend writes part 4 as a nested envelope (`{legendSettings, exportOptions, publishState, eatOverlayEnabled, eatConfidenceThreshold}`); Python expected the flat annotation-keyed shape and called `.get("categories")` on every top-level value → `AttributeError` on the EAT bool. `ArrowReader` swallows it, so `protspace serve`/`style` silently showed **every legend colour and shape as empty**.

Added `unwrap_settings`/`rewrap_settings` — the Python half of the contract `normalizeBundleSettings` already implements. Restyling a frontend export now preserves its `publishState`/EAT settings instead of flattening them. The envelope is detected the same way the frontend's own `isNormalizedBundleSettings` does it (a `legendSettings` map plus an envelope sibling), so a flat map whose annotation column happens to be named `legendSettings` is not misread.

## 3. Numeric type identity lost on export (`1d0326e3`, `614cb420`, `d0df0d28`)

- Numeric columns were hardcoded to parquet `DOUBLE`. Python keys legends/styles off `str(value)`, so `'100'` became `'100.0'` and a valid `protspace style` template started failing validation. The physical type is now chosen from the annotation's declared type: **INT32** for integers (values passed through untouched — no per-protein bigint array), INT64 beyond the int32 range, DOUBLE otherwise.
- Kind and int/float identity weren't persisted, so a numeric column with no surviving values (isolation mode / query filter) reloaded as categorical with one `__NA__` category. The reader now derives that identity from the annotations part's **parquet schema** — no bespoke metadata key, and it survives Python's rewrite paths, which drop key-value metadata but preserve column types.
- Only **integer** physical types are authoritative. Bundles exported before this writer stored every numeric column as DOUBLE, so DOUBLE carries no int/float information and must not re-label their integer annotations; those keep pure value inference, exactly as before. Columns stored as text — the CLI stringifies its annotation frame — carry no numeric type at all and are unaffected.

## 4. All-missing numeric column rendered blank (`a5d8f110`)

`materializeNumericAnnotation`'s zero-value branch left every row on the `-1` sentinel — empty legend, every point hidden. It now emits the N/A pseudo-bin the populated path already appends. Pre-existing for genuinely all-null numeric columns; fix 3 routes more columns here, so it had to go first.

## Deliberately out of scope

Two further confirmed export losses are already being handled on other branches and are untouched here: 5-part `--stats` bundles being rejected on read (#389) and the missing delimiter-collision guard (#366). The 21 pre-existing ESLint warnings are split out as #395 — this PR touches zero `apps/web` files.

## Verification

- `pnpm test`: 6/6 tasks — utils 322, core 1394, app 165, backend 794
- `pnpm format:check`, `pnpm type-check`, `pnpm lint` (0 errors), `ruff check` + `ruff format --check` all clean
- Every fix has a regression test, each confirmed to **fail** against the unfixed code before being kept
- Reviewed by adversarial multi-agent passes at three points. They caught, and this PR fixes: a Prettier failure `lint-staged` had hidden, a `legendSettings`-named-annotation collision, a blank-plot regression, a one-sided wire-format key that the parquet schema already encoded, and the DOUBLE re-labelling above. The last was checked by measuring the real fixtures — `5K.parquetbundle`, `phosphatase_no_binning` and `data_custom` store every annotation as text, so they were never exposed.

https://claude.ai/code/session_01GJx5dB9pA6MjdCjEw4ytW2


t03i:
## Code review (high effort, automated multi-agent + verification pass)

24 agents, 22 candidates, 10 refuted, **8 verified findings**. Two clusters dominate:

- **(a)** the physical parquet type is now the sole carrier of numeric identity — it misreads pyarrow's all-null `INT32` columns and makes the int/float round trip one-directional;
- **(b)** the `__NA__` sentinel drop leaves the Python styling path with no spelling for a missing cell.

---

### 1. `packages/core/src/components/data-loader/utils/bundle.ts:81` — all-null pyarrow column reads as INT32 (CONFIRMED)

`readNumericColumnTypes` treats the `INT32` physical type pyarrow uses for arrow-`null` (all-missing) columns as an integer declaration, so `restoreDeclaredNumericAnnotations` (`conversion.ts:686`) converts all-missing **categorical** columns from Python-written bundles into numeric annotations.

A pandas column that is entirely `None` (e.g. `ec_number` when no protein got a hit) becomes arrow type `null`, which `pq.write_table` stores as `optional int32 ec_number (Null)` — physical type `INT32`, logical type `Null`, which the new code does not inspect. The legend/filter UI then offers `>`/`<`/`between` instead of the categorical value list, and re-exporting from the browser writes the column back as a real `INT32` numeric column — permanently converting a text annotation to numeric.

### 2. `apps/protspace/src/protspace/data/io/settings_converter.py:385` — `rewrap_settings` drops numeric annotations' `legendSettings` (CONFIRMED)

`rewrap_settings` replaces the whole `legendSettings` map with the rebuilt one, but `visualization_state_to_settings` only rebuilds annotations present in `annotation_colors`/`marker_shapes`. Numeric annotations never are: the frontend persists their categories with `color: ""`/`shape: ""`, and `settings_to_visualization_state` filters empty colors/shapes out.

`protspace style --styles family.json` on a bundle with categorical `family` + numeric `plddt` loses `plddt`'s `numericSettings` (binning strategy, signature, `reverseGradient`), `selectedPaletteId`, `hiddenValues` and `maxVisibleValues` — exactly the non-lossy-rewrite property this PR set out to add. Same root cause at `:235` and `:123`.

### 3. `packages/utils/src/parquet/bundle-writer.ts:143` — NULL-on-write leaves Python no way to style the N/A group (CONFIRMED)

Missing categorical cells now export as parquet NULL. `ArrowReader.get_all_annotation_values` returns `None`, and `compute_value_frequencies` keys the group as `str(None)` = `"None"`. A style file naming the missing group — `__NA__`, exactly what `protspace style --template` emitted for the same bundle before this change, or `<NA>` — hits `add_annotation_style.py:283` where `_resolve_na` fails (`_NA_LABELS = {"", "<NA>", "NaN"}` contains neither `__NA__` nor `None`), aborting with `ValueError: Value '__NA__' does not exist for annotation 'family'`. Styling the literal `"None"` succeeds, but `settings_converter.py:368` only rewrites `_NA_LABELS` members to the `__NA__` key, so the colour persists under `categories["None"]` and the web app's `categories["__NA__"]` swatch silently stays gray.

### 4. `packages/utils/src/parquet/bundle-writer.ts:74` — INT32 write changes value keys `'100.0'` → `'100'` (CONFIRMED)

`compute_value_frequencies` (`add_annotation_style.py:130`) and `_annotation_display_values` (`:151`) key every annotation value off `str(v)` read straight from the parquet column, and `add_annotation_styles_bundle` validates each style key against that set. A bundle exported by an earlier build stored `length` as `DOUBLE`, so existing style YAMLs and persisted `categories` maps use `'100.0'`. After re-export the column is `INT32` and the keys become `'100'`: `protspace style -s styles.yaml` aborts with `ValueError: Value '100.0' does not exist for annotation 'length'`, and stored per-value colours silently stop applying. Nothing in the diff migrates the old keys.

### 5. `packages/utils/src/parquet/bundle-writer.ts:143` — `isNAValue` guard deletes a real category spelled `__NA__` (CONFIRMED)

`normalizeMissingValue` only nulls out `MISSING_VALUE_TOKENS` (`na`, `n/a`, `nan`, `null`, `none`); the literal string `__NA__` is not in that set, so a CSV containing a category spelled `__NA__` is ingested as a genuine, palette-assigned category. The previous writer round-tripped it. The new unconditional `isNAValue(value)` guard writes every such cell as NULL: on reload the category is gone from the legend, the proteins render with the N/A swatch, and `protspace stats` no longer counts the class.

### 6. `apps/protspace/src/protspace/data/io/settings_converter.py:89` — `_is_frontend_envelope` can misclassify a flat settings map (PLAUSIBLE)

The sibling-key check only needs one member of `_ENVELOPE_SIBLING_KEYS`. Annotation names come from user CSV headers, so a flat, Python-written settings map for a dataset whose columns include both `legendSettings` and any of `exportOptions`/`publishState`/`eatOverlayEnabled`/`eatConfidenceThreshold` is misclassified as an envelope. `unwrap_settings` then returns the `legendSettings` column's own entry, `settings_to_visualization_state` iterates its inner keys (`categories`, `sortMode`, …) and produces an empty `visualization_state` — so `protspace style` writes back a part-4 with every annotation's saved colours and shapes dropped, no warning. That is the exact data loss the guard's docstring says it prevents, just requiring two colliding column names instead of one.

### 7. `docs/guide/data-format.md:175` — doc says `DOUBLE` is authoritative; the reader ignores it (CONFIRMED)

`conversion.ts:675` only honours `INT32`/`INT64` and lets value inference win for `DOUBLE` (the branch's own test "lets inference decide int/float for a DOUBLE column" and the `raw_numeric_test.parquetbundle` fixture pin that). A maintainer reading this bullet as the spec would "fix" the reader to trust `DOUBLE` as `float`, re-labelling every pre-INT32 export's integer annotation and turning legend labels from `10 - 25` into `10.0 - 25.0` — the regression the code comments explicitly guard against.

### 8. `packages/core/src/components/data-loader/utils/conversion.ts:664` — numeric rescue is a post-hoc repair, not fed into inference (PLAUSIBLE)

For an all-missing numeric column the categorical builder still runs to completion — unique-value scan, palette colour/shape allocation, synthetic `__NA__` category, full per-protein `annotation_data` store — and this pass then deletes all of it and allocates a *second* protein-count array of nulls, even though `inferAnnotationType` already produced exactly that array as `inference.numericValues` and discarded it (`conversion.ts:1243`, `:1482`). At Swiss-Prot scale that is real wasted allocation, and the int/float rule now lives in four places, so a future converter added without a matching `.then(restoreDeclaredNumericAnnotations)` silently regresses the fix.

t03i:
@tsenoner all eight review findings above are fixed in abe27bcd — ready for another look.

| # | Finding | Fix |
|---|---------|-----|
| 1 | pyarrow all-null column reads as INT32 | `readNumericColumnTypes` ignores any field carrying a logical/converted type (NULL, DECIMAL, …); new pyarrow fixture `all_null_column.parquetbundle` pins it |
| 2 | `rewrap_settings` drops numeric annotations | overlays the rebuilt map on the template's `legendSettings` instead of replacing it |
| 3 | Python cannot style the N/A group of NULL cells | `"None"` and `"__NA__"` joined the NA aliases on both sides |
| 4 | `'100.0'` → `'100'` orphans style keys | style keys resolve numerically as a fallback (`_resolve_numeric`) |
| 5 | a real category spelled `__NA__` was deleted on export | the spelling is reserved at ingestion (`MISSING_VALUE_TOKENS`) |
| 6 | `_is_frontend_envelope` misclassifies a colliding flat map | the inner shape settles it — an annotation-keyed map never carries per-annotation settings keys at its top level |
| 7 | doc claims `DOUBLE` is authoritative | bullet rewritten to say only integer physical types are, and why |
| 8 | numeric rescue is a post-hoc repair pass | **not applied** — see below |

Finding 8 (`restoreDeclaredNumericAnnotations` re-doing work `inferAnnotationType` already did) is a real observation but it is a wash-behaviour refactor across three converters on the hot load path, for no output change. Left as is deliberately; happy to take it in a follow-up if you'd rather have the rule in one place.

Checks: 799 Python tests + 1721 TS tests pass, ruff/eslint/tsc clean. Each fix has a regression test — the pyarrow one was verified to fail without the guard.

PR #398, opened 2026-08-01
fix(search): mark already-selected proteins in search suggestions

### LINKED ISSUE #345: [FEATURE] Searching for a point already in the selection should give a constructive message
### Feature Request

When searching for a protein that is already selected instead of showing "No matching protein IDs found" something like "Protein ID is already selected" should be shown.

### Code of Conduct

### LINKED ISSUE #413: Search: arrow-key highlight scrolls out of view in the suggestion dropdown
## Summary

Arrow-key navigation in the protein search dropdown moves the highlight past the visible area without scrolling it into view. Once the highlight passes the last visible row (~8 rows at the current `max-height`), the user is navigating blind — the highlighted row is off-screen and the scrollbar does not follow.

## Reproduction

1. Open `/explore` with any dataset that has more than ~9 IDs matching a query.
2. Focus the search box (or type a short prefix) so the suggestion dropdown opens.
3. Press <kbd>↓</kbd> repeatedly.
4. After the 8th or 9th press the highlight leaves the visible area. The list does not scroll, so no row appears highlighted, and pressing <kbd>Enter</kbd> activates a protein the user cannot see.

## Cause

`.search-suggestions` is a fixed-height scroll container (`packages/core/src/components/control-bar/search.styles.ts`):

```css
.search-suggestions {
  max-height: 20rem;
  overflow-y: auto;
}
```

`_onSearchKeydown` in `packages/core/src/components/control-bar/search.ts` updates `highlightedSuggestionIndex` for <kbd>↓</kbd>/<kbd>↑</kbd> but nothing ever scrolls the newly highlighted row into view.

## Why it surfaced now

This was previously unreachable. `_flushSuggestions()` recomputed suggestions on every keypress and reset the highlight to 0, so repeated <kbd>↓</kbd> was stuck at index 1 and could never reach the fold. Fixing that (making the flush conditional on a pending debounce) is what exposed this.

## Suggested fix

Scroll the active row into view whenever `highlightedSuggestionIndex` changes, using `scrollIntoView({ block: 'nearest' })` so the list only scrolls when the row is actually outside the viewport and the page itself never jumps.

## Related

Found while reviewing #398.

FlorinSenoner:
## Summary

- keep already-selected proteins in the suggestion list, rendered as marked rows, instead of filtering them out
- activating a marked row removes that protein (`remove-selection`), making search a toggle
- delete the empty-state message classification entirely — `No matching protein IDs found` is now truthful
- fix the latent bugs this work exposed, plus #413

## Root cause

`computeSearchSuggestions` dropped already-selected IDs from the suggestion list. When every prefix match was already selected the list came back empty, and the renderer treated an empty list as "nothing matched". With P00595–P00599 selected, searching `P0059` reported no matches for five matching proteins. The message was false, not merely unhelpful.

This PR originally classified that empty state (`Protein ID is already selected` when the query exactly matched a selected ID). That covered only exact whole-ID queries with a completely empty list, so two reachable states stayed wrong: a partial query resolving only to selected proteins still misreported, and a selected ID that is a strict prefix of unselected ones (`GT4` alongside `GT40`–`GT46`) produced a non-empty list and therefore no feedback at all.

Keeping selected proteins in the list removes the distinction entirely — there is one rule, and no render-time classification.

## What changed

- **`search-suggestions.ts`** — returns `SearchSuggestion[]` (`{ id, isSelected }`) instead of `string[]`. Selected and selectable entries draw from independent budgets (`MAX_SELECTED_SUGGESTIONS` = 10, `MAX_SEARCH_SUGGESTIONS` = 50) so addable proteins stay visible however many matches are selected. Natural array order, interleaved.
- **`search.ts`** — marked rows with `aria-selected` / `aria-label`; activation toggles; new `remove-selection` event; explicit `isSuggestionDropdownOpen` state; `_getEmptyStateMessage()` deleted.
- **`control-bar.ts`** — `_handleSearchSelectionRemove` mirrors `_handleSearchSelectionAdd`, minus the structure-viewer call.

Removal preserves the query and the open dropdown; adding still clears both, preserving the search-one-add-one flow.

## Bugs fixed along the way

| Bug | Impact |
| --- | --- |
| Early-exit required *both* budgets full | With <10 selected matches it never fired — 573,000 of 573,000 index reads per recompute on Swiss-Prot scale. Now 50. |
| Early exit bounded by the *highest index* of any selected ID | Even after the above, the scan ran until it had walked past every selected ID — an arbitrary position, since selections come from scatter-plot clicks. See **Scan bound** below. |
| A bare focus armed a destructive **Enter** | `_onInputFocus` seeded the cursor at row 0, and on an empty query row 0 is the current selection rendered as a marked row. Clicking into the box and pressing Enter dropped a protein the user never pointed at — harmless before marked rows existed, destructive once activation became a toggle. |
| `aria-expanded` collapsed while the popup was visible | The combobox reported itself closed during `No matching protein IDs found`, and `aria-controls` named nothing: a screen-reader user was told the opposite of what sighted users saw, and the message was never announced. |
| A duplicated selected ID rendered twice | Two identical marked rows, each claiming to remove the same protein; the second click a no-op against an already-removed ID. |
| `_flushSuggestions()` reset the highlight on every keypress | Arrow navigation was stuck at index 1 and Enter always activated row 0 — destructive once Enter can remove. |
| `willUpdate` re-opened a *closed* dropdown | ~50 unrelated rows popped open after every successful add; Escape didn't stick. |
| Highlight scrolled out of view (#413) | Past ~8 rows the user navigated blind. Fixed with `scrollIntoView({ block: 'nearest' })`. |
| Result count used as a proxy for "dropdown is open" | An open no-match state also has zero rows, so a dataset change could not refresh it. Now an explicit flag. |
| `_onSearchInput` opened the dropdown before its results existed | The first keystroke after any add/Escape/blur rendered the *previous* close's empty list against a non-empty query, flashing `No matching protein IDs found` for the full 120 ms debounce — the exact false message this PR removes. Opening now happens with the results. |
| Suggestion clicks reopened the dropdown they closed | The list is a DOM child of `.search-container`, whose `@click` refocused the input. Handler moved to `.search-chips`. |
| A preserved highlight of -1 was promoted to row 0 | A dataset-driven refresh invented a keyboard cursor the user never moved, sending Enter to an arbitrary match. |
| `close-search` registered as an inline closure | Unremovable, so every re-attach stacked another handler and leaked the detached element. |

## Scan bound

`computeSearchSuggestions` stopped only once the selectable budget was full **and** every selected ID had been walked past. The second clause is bounded by the *highest index* of any selected ID, and selections come from clicking scatter-plot points — so that index is arbitrary. Expected scan length for k selections was `N·k/(k+1)`: half the array for one selection, ~90% for three, paid per debounced keystroke plus every Enter/Arrow flush and every selection echo.

Counting up front how many selected IDs the query can actually match (O(|selection|), capped at the budget) gives the loop a target it can *reach*, so it now stops at the budgets. Measured on a 573K counting `Proxy`, typing a query that matches no current selection:

| | before | after |
| --- | --- | --- |
| 1 selection | 400,001 index reads | **50** |
| 3 selections | 500,001 index reads | **50** |

Ordering is untouched — marked rows still come out in `availableIds` order, because the scan still finds them there. This is deliberately *not* the alternative the earlier docstring floated (break on the selectable budget alone and top up from `selectedIds`), which reorders overflow marked rows by selection rather than by `availableIds` index and so contradicts the spec.

Verified by differential fuzz against the previous implementation: 40,000 randomized `(ids, selection, query, limits, Set-vs-array)` configurations, **0 mismatches** on duplicate-free input. The only divergence is a duplicated selected ID in `availableIds` — which the previous code emitted once per *occurrence*, and only depending on the arbitrary positions of unrelated selected IDs. It is now one row per protein. Both preconditions on `availableIds` are stated in the docstring.

The per-entry budget check also moved above the prefix test, dropping a throwaway `toLowerCase()` on every entry the scan passes while the other budget fills.

## Cleanup

Reviewed for reuse/simplification/efficiency/altitude, then applied:

- **`_commitSelection()`** in `control-bar.ts` — the assign-chips / assign-count / mirror-to-scatterplot / dispatch block was copied verbatim across six handlers and this PR had added a seventh. Routing `handleClearSelections` through it also fixed a latent bug: it reset `selectedProteinsCount` *inside* the auto-sync branch, so clearing while `autoSync` was false (as `data-renderer.ts` sets it during a data swap) emptied the chips but left the count stale, keeping the Clear button live.
- **`scrollHighlightedIntoView()`** in `dropdown-helpers.ts` — three copies existed and the newest had already diverged, omitting `behavior: 'smooth'`, so protein search jumped where the sibling dropdowns glided.
- **`_closeDropdown()` / `_resetSearch()`** — adding the open flag left the same four-line close block across seven sites.
- **`render()`** — a triple-nested ternary spanning 42 of its 73 lines, re-deriving the listbox condition `_isListboxRendered` already expressed, so the ARIA and the DOM could drift apart silently. Split into `_renderSuggestions()` (guard clauses) and `_renderSuggestion()`.
- **Combobox ARIA** — the input is now a `role="combobox"` owning the listbox, with `aria-expanded`, `aria-controls`, and `aria-activedescendant` naming the highlighted row. #413 fixed the sighted half of navigating blind; the keyboard cursor was announced nowhere. With the cursor carried by `aria-activedescendant` and the list marked `aria-multiselectable`, `aria-selected` now carries its real listbox meaning instead of doubling as the cursor. The no-match state is the same popup — same id, same `role="listbox"` — so `aria-controls` always names a real one.

## Also fixed here (pre-existing, not introduced by this work)

Three things this review turned up in the files it was already touching. Each is its own commit, so any of them can be dropped.

- **Stacked dropdown listeners** — `annotation-opened` / `search-opened` in `control-bar.ts` and `close-dropdown` in `annotation-select.ts` were registered as inline closures, so every re-attach stacked another handler and pinned the detached element. That is the identical defect this PR fixes for `close-search` in `search.ts`; `annotation-select.ts` had no `disconnectedCallback` at all. Fixing one of the three and leaving the siblings seemed worse than the small scope increase.
- **Dead selected-protein caption** — `getElements()` resolved `#selectedProtein` and threaded it through `runtime.ts` into the interaction controller, where `updateSelectedProteinDisplay()` wrote a caption into it from four call sites. No element with that id exists anywhere in the repo, and `git log -S 'id="selectedProtein"' --all` finds no commit that ever added one — so the lookup was always `null` and every call returned on the first guard. Removed. The caption users actually see belongs to the scatterplot.
- **`_addSelection` rescanned the dataset twice** — `includes(id)`, then `find((p) => p.toLowerCase() === id.toLowerCase())`, which recomputed the loop-invariant `id.toLowerCase()` once per available protein — 573K throwaway strings per call on top of the 573K the other side already costs. Extracted `_canonicalProteinId()` with the invariant hoisted. Deliberately *not* indexed with a cached lowercase `Map`: that would be O(1) but holds ~573K new strings for the session and would have to build on the first suggestion click, trading a sub-millisecond `includes` for a multi-hundred-millisecond build. This is a per-add path, not per-keystroke.

Two claims raised in review were investigated and **refuted**, so nothing changed for them: per-row `@mousedown` closures do *not* cause listener churn (lit-html's `EventPart` registers itself as the listener and only swaps the stored function — verified in `lit-html.js:1306-1366` and empirically at 60 rows × 10 full content swaps → 0 `removeEventListener`), and routing the isolation handlers through `_commitSelection` fixes no reachable symptom while widening what `protein-selection-change` means.

## Behaviour changes

- Pressing Enter on an already-selected ID now **removes** it (previously a no-op).
- Focusing the empty input now lists current selections alongside selectable entries, and **no longer seeds a keyboard cursor** — Enter does nothing until an arrow key or a typed query establishes one.
- Arrow navigation advances past the first row, activates the highlighted row on Enter, and scrolls it into view.

## Known limitation

Repeated **Enter** on marked rows does not prune successive proteins. The first Enter removes the highlighted protein; the row stays at that index but flips to selectable, and the highlight is preserved by index — so a second Enter re-adds what the first removed and clears the query. Verified:

```
REMOVED: ["P00595"]  ADDED: ["P00595"]   QUERY AFTER: ""
```

Pruning several from one result set works by **clicking**, and by arrowing between keystrokes. Making repeated Enter walk the list is a toggle-semantics decision left out of this PR. The deeper fix is to carry the keyboard cursor as a protein ID rather than as an index into a list that is recomputed underneath it, which would also retire `preserveHighlight` and its clamp.

## Verification

- full JS suite: 2,032 passing, 1 pre-existing skip (`@protspace/core` 1,499, `@protspace/utils` 368, `@protspace/app` 165)
- `pnpm knip` / `knip --dependencies`: clean (no exports orphaned by the dead-code removal)
- `pnpm format:check`, `pnpm lint` (0 errors), `tsc --noEmit`, `pnpm precommit`: clean
- `openspec validate --specs`: 5 passed, 0 failed — the change is **archived in this PR** as
  `openspec/changes/archive/2026-08-10-mark-selected-proteins-in-search/`, and its four
  requirements are merged into the new `openspec/specs/protein-search-feedback/`. AGENTS.md now
  requires archiving as the last commit on the branch rather than after the merge, since a
  closed PR and a deleted branch mean it never happens and `openspec/specs/` silently drifts.
- iteration-bound regression tests (Proxy index-read counters, not timing) pin the early exit — including the case where the query cannot match any selected ID, which the previous test missed by pinning a selection at index 2, the best case
- differential fuzz of the scan against the pre-change implementation, 40,000 configurations
- browser-verified on `/explore`: a query whose every match is selected lists them all marked with no false message; clicking a marked row removes it in place with the query preserved; a query matching nothing still reports no matches; arrow navigation scrolls the highlight into view in both directions

No `apps/protspace/**` files are touched, so this merges without cutting a PyPI release; it does trigger the GitHub Pages deploy.

Closes #345
Closes #413


tsenoner:
### Automated review

**Does it solve #345?** Partially — the wiring is real and the new jsdom test drives the actual custom element through its input and debounce, so typing a full ID that is already selected now shows "Protein ID is already selected" instead of the misleading text. But the message needs both a whole-ID query and a completely empty suggestion list, so two reachable states still misbehave: a partial query that narrows to only an already-selected protein still renders "No matching protein IDs found", and when the selected ID is a strict prefix of other IDs the box never renders at all. On the reporter's 149-ID dataset both shapes occur (31 IDs have a unique-resolving strict prefix; GT4/GT5/GT9/GT10-GT13/GT15 are shadowed by GT40-GT46, GT90-GT99, ...).

Found 3 issues:

1. The already-selected message is gated on an exact-ID query *and* an empty suggestion list, so the misleading string survives for partial queries and no feedback appears at all for prefix-colliding IDs — suggestions are built by prefix match with selected IDs dropped (`search-suggestions.ts` L26-L27), so with D7U9A6 selected typing `d` empties the list and still says "No matching protein IDs found"; select GT4 and type `GT4` and the dropdown lists GT40-GT46 while silently omitting GT4. One rule covers both: when every available ID that prefix-matches the query is already selected, say so — or render the note above the suggestion list rather than only in place of it.

https://github.com/tsenoner/protspace/blob/d1cb2c10cb250bf11a74c2aa05f0f332ca0aabe3/packages/core/src/components/control-bar/search.ts#L69-L77

2. `_getEmptyStateMessage()` scans the entire available-ID array with a per-element `toLowerCase()` from inside `render()`, so it re-runs on every keystroke while the no-results box is open — `searchQuery` is `@state` and assigned synchronously in `_onSearchInput` (only `_updateSuggestions()` is debounced), and `control-bar.ts:624` binds the full `data.protein_ids` array (~573K for Swiss-Prot, ~5.6 ms per scan measured warm), reintroducing the per-keystroke O(N) work that 8048bf4a removed. Cheapest exact-behaviour fix: return the generic message immediately when `selectedProteinIds.length === 0`, then look the query up in `selectedProteinIds` first and only confirm against `availableProteinIds`.

https://github.com/tsenoner/protspace/blob/d1cb2c10cb250bf11a74c2aa05f0f332ca0aabe3/packages/core/src/components/control-bar/search.ts#L221-L233

3. The proposal's Impact section claims the change "adds a narrow browser regression to the existing `apps/web` Playwright suite", which the diff does not contain — no file under `apps/web` is touched, the only test added is `search.component.test.ts`, and `design.md` L35 explicitly decides the opposite. Since archived changes keep proposal.md verbatim, this will persist as a false claim about shipped coverage; dropping or rewording the bullet is a one-line fix.

https://github.com/tsenoner/protspace/blob/d1cb2c10cb250bf11a74c2aa05f0f332ca0aabe3/openspec/changes/show-already-selected-search-message/proposal.md#L23-L27

<sub>Reviewed at d1cb2c10cb250bf11a74c2aa05f0f332ca0aabe3 against issue #345.</sub>

FlorinSenoner:
Independent triage at d1cb2c10cb250bf11a74c2aa05f0f332ca0aabe3:

1. **Split disposition.** The partial-prefix case is non-actionable for this scoped change: the OpenSpec design explicitly makes partial-match semantics a non-goal, and the requirement targets exact IDs. The exact-ID prefix-collision case is actionable, though: when selected `GT4` coexists with unselected `GT40`-`GT46`, the suggestions branch wins and the exact selected-ID feedback promised by the proposal/design/scenario is never rendered. Required direction: show exact selected-match feedback even when other selectable prefix suggestions remain, without making the selected ID actionable.
2. **Actionable.** `_onSearchInput` updates reactive `searchQuery` immediately, while only suggestion recomputation is debounced. Whenever the current suggestion list is empty, render calls `_getEmptyStateMessage()`, whose `availableProteinIds.find(...toLowerCase())` can scan the full list on each keystroke. That conflicts with the large-dataset performance intent of `8048bf4a`. Required direction: check selected IDs first and confirm availability only for a candidate, or use an equivalent cached normalized lookup.
3. **Actionable.** The diff adds only `search.component.test.ts`; no `apps/web` Playwright test is added, and design line 35 explicitly chooses manual browser verification. The proposal Impact bullet should be corrected to describe component coverage plus manual browser verification.

Follow-up implementation is needed for the exact-ID collision case, the render-time lookup, and the proposal correction. I would not expand this change to partial-prefix messaging.

FlorinSenoner:
Verified the actionable review feedback is addressed on current head `d857eaedb94d1cc06ee70ac30ba7662927dfb3c8`:

- Selected IDs now remain in the prefix result set as marked rows, including all-selected partial queries and the `GT4` / `GT40` prefix-collision case; activating a marked row removes it.
- `_getEmptyStateMessage()` and its render-time full-array scan are gone. Focused iteration-count regressions also verify early exit once suggestion budgets are satisfied.
- The OpenSpec proposal now accurately records component coverage plus manual browser verification and explicitly states that no Playwright test was added.

Fresh verification: 47 focused tests passed; `openspec validate mark-selected-proteins-in-search --strict`, `pnpm format:check`, the full `pnpm test:ci` suite (1,884 passed, 1 pre-existing skip), and `pnpm precommit` all completed successfully. No later review comments or inline threads were present.

docs/superpowers/plans/2026-08-04-search-selected-marking.md
```diff
@@ -0,0 +1,1231 @@
+# Mark Selected Proteins in Search Suggestions — Implementation Plan
```
FlorinSenoner: [P2] Remove the force-added duplicate planning artifacts

AGENTS.md says plans, specs, and designs must live under openspec/changes and explicitly forbids docs/, while .gitignore labels docs/superpowers as “kept local, never committed.” This PR force-adds this 1,231-line plan plus the 281-line sibling design, duplicating the canonical OpenSpec record; the copy has already drifted (line 49 says search.test.ts is untouched although this diff deletes 76 lines there). Please remove both docs/superpowers additions and keep the change record only under openspec/changes/mark-selected-proteins-in-search/.

openspec/changes/mark-selected-proteins-in-search/specs/protein-search-feedback/spec.md
```diff
@@ -0,0 +1,51 @@
+## ADDED Requirements
+
+### Requirement: Protein search shows already-selected proteins as marked suggestions
+
+The protein search SHALL include already-selected protein IDs in the suggestion list, visually marked as selected, rather than omitting them. Selected and selectable entries SHALL draw from independent display budgets so that selectable entries remain visible regardless of how many matches are already selected.
+
+#### Scenario: Every prefix match is already selected
+
+- **WHEN** a user searches a partial ID whose every available match is already selected
+- **THEN** each matching protein is listed and marked as selected
```
FlorinSenoner: [P2] Make the scenario honor the selected-row cap

This requires each matching selected protein to be listed, but computeSearchSuggestions deliberately stops after MAX_SELECTED_SUGGESTIONS = 10, and the new unit test asserts that cap. With 11 selected IDs sharing the query prefix, one match is omitted, so the implementation violates this SHALL scenario. Either implement the stated all-match behavior or revise the requirement and scenario to define the capped behavior, including what the user sees when additional selected matches are hidden.

packages/core/src/components/control-bar/search.ts
```diff
@@ -100,6 +107,39 @@ class ProtspaceProteinSearch extends LitElement {
...
+    // `searchSuggestions` is computed on a debounce, so a selection change made elsewhere
+    // (including this component's own remove, which keeps the query) would otherwise leave
+    // the open dropdown stale. Recomputing here cannot loop: it only writes
+    // `searchSuggestions`, never `selectedProteinIds`.
+    if (!changed.has('selectedProteinIds') && !changed.has('availableProteinIds')) return;
+    if (!this.searchQuery.trim() && !this.isInputFocused) return;
+    // Only refresh a dropdown that is already open. Without this, clearing the query
+    // on add (or Escape, or a blur that left text behind) leaves the component ready
+    // to re-open the list the moment the parent echoes a selection change back down.
+    if (this.searchSuggestions.length === 0) return;
```
FlorinSenoner: [P2] Recompute open empty results when the dataset changes

searchSuggestions.length === 0 does not mean the dropdown is closed: render() uses that exact state with a nonempty query to show “No matching protein IDs found” at lines 77-82. If query NEW has no match in the old availableProteinIds and the parent then swaps to a dataset containing NEW1, this guard ignores the watched availableProteinIds change and leaves false no-match feedback. Please track open or closed state separately, or let availableProteinIds changes recompute while retaining the add and Escape reopen protection, and cover the empty-result dataset-swap case.

docs/superpowers/plans/2026-08-04-search-selected-marking.md
FlorinSenoner: Fixed in `a668f401`: removed both force-added `docs/superpowers/` planning artifacts (1,512 lines). The canonical design/spec/task record now remains only under `openspec/changes/mark-selected-proteins-in-search/`. Focused tests (48/48), strict OpenSpec validation, and `pnpm precommit` passed.

openspec/changes/mark-selected-proteins-in-search/specs/protein-search-feedback/spec.md
FlorinSenoner: Fixed in `a668f401`: the requirement now explicitly caps selected suggestions at 10 rows, and the scenario specifies the first 10 matches in available-protein order with additional selected matches not rendered. Runtime work remains bounded. Strict OpenSpec validation and `pnpm precommit` passed.

packages/core/src/components/control-bar/search.ts
FlorinSenoner: Fixed in `a668f401`: dropdown openness is now explicit instead of inferred from result count. Open empty results recompute when `availableProteinIds` changes, while add, Escape, and blur keep the dropdown closed. The regression was observed failing before the fix and now passes within the 48/48 focused tests; `pnpm precommit` also passed.

docs/superpowers/plans/2026-08-04-search-selected-marking.md
FlorinSenoner: Revalidated on current head f39e3e07: the complete PR diff contains no docs/superpowers files; the canonical proposal, design, spec delta, and tasks remain only under openspec/changes/mark-selected-proteins-in-search/. Focused affected suites pass 52/52, strict OpenSpec validation passes, and a fresh pnpm precommit completes successfully.

openspec/changes/mark-selected-proteins-in-search/specs/protein-search-feedback/spec.md
FlorinSenoner: Revalidated on current head f39e3e07: the requirement fixes the selected-entry display budget at 10 rows, and the all-selected scenario says that only the first 10 matches in available-protein order are rendered while overflow remains hidden. The focused cap regressions pass within 52/52 affected tests; strict OpenSpec validation and fresh pnpm precommit also pass.

packages/core/src/components/control-bar/search.ts
FlorinSenoner: Revalidated on current head f39e3e07 after the later dropdown-lifecycle commits: openness is still tracked explicitly by isSuggestionDropdownOpen. The regression covers an open empty result refreshing when availableProteinIds changes, then verifies an Escape-closed dropdown stays closed across a later dataset change. Focused affected suites pass 52/52 and fresh pnpm precommit completes successfully.

PR #399, opened 2026-08-01
fix(protspace): preserve unlabeled TED domain names

### LINKED ISSUE #342: [FEATURE] Naming of TED domains without CATH label
### Feature Request

TED domains without CATH labels, showing up on the TED website as "-", should also show up as "-" in ProtSpace instead of "unclassified". Example for such a protein: https://ted.cathdb.info/uniprot/W6JQJ9

### Code of Conduct

FlorinSenoner:
## Summary

- preserve TED's `-` label for domains without a CATH assignment
- retain pLDDT and existing multi-domain serialization
- add regression coverage and align generated/package annotation documentation
- document the behavior through the OpenSpec change `fix-ted-unlabeled-domain-name`

## Root cause

`TedRetriever._format_domains` explicitly replaced a missing or `-` CATH label with the literal
`unclassified`. Because this substitution happened at the backend serialization boundary, every
downstream consumer received a label that differed from TED.

## Reproduction

Using W6JQJ9 against the AlphaFold domains API:

1. Fetch `https://alphafold.ebi.ac.uk/api/domains/W6JQJ9`.
2. Observe domain 2 has `cath_label: "-"` and pLDDT `96.7064`.
3. On main, format that payload and observe `unclassified|96.7`.
4. With this change, the same payload formats as `-|96.7`, while labeled domains and ordering
   remain unchanged.

## Tests

- TDD red: focused regression failed with expected `-|90.5`, actual `unclassified|90.5`
- TDD green: focused regression, 1 passed
- TED retriever module: 9 passed
- Python CI-equivalent suite: 787 passed, 6 deselected
- Ruff check and format check: passed
- OpenSpec strict validation: passed
- `pnpm precommit`: passed before commit and push

Closes #342


tsenoner:
### Automated review

**Does it solve #342?** Yes — the one-line change in `TedRetriever._format_domains` makes unlabeled TED domains emit `-|{plddt}` instead of the invented `unclassified`, matching ted.cathdb.info, and the test is tightened to an exact `== "-|90.5"` assertion. I traced the new literal across the whole boundary: `-` is absent from both missing-value vocabularies (`core/constants.py` and `packages/utils/src/visualization/missing-values.ts`), `_strip_scores_from_cell` under `--no-scores` yields a bare `-`, and `parseAnnotationValue` resolves `-|96.7` to label `-` with score 96.7, so it renders as a real legend category rather than being swallowed as N/A. Docs are updated consistently on both the web and Python sides, and no `unclassified` source literal remains.

Found 1 issue:

1. Re-running `protspace prepare` into an existing output directory keeps the old `unclassified` label — the cache stores the *formatted* `ted_domains` strings, so when `missing` is empty the pipeline returns the cached frame and the patched formatter never runs. `prepare` defaults to `keep_tmp=True` with `cache_dir = output_dir / "tmp"` and a default output of `.`, so anyone verifying this fix in the same working directory still sees the old value, with no message explaining why. The escape hatch (`--refetch ted`) exists but nothing ties it to this change: the PR body, both annotation docs, and `design.md` say only "No stored-data migration is required." A one-line note in the annotation docs or the change's migration plan would close this.

https://github.com/tsenoner/protspace/blob/a07547f1cb94552bca6cb8fb41d9b82082c1368b/apps/protspace/src/protspace/data/processors/pipeline.py#L410-L436

<sub>Reviewed at a07547f1cb94552bca6cb8fb41d9b82082c1368b against issue #342.</sub>

FlorinSenoner:
Disposition: actionable; the cache caveat needs to be documented. Verified at `a07547f1cb94552bca6cb8fb41d9b82082c1368b`: `prepare` defaults to `output=Path(".")` and `keep_tmp=True`, stores formatted annotations in `{output}/tmp/all_annotations.parquet`, and `_fetch_annotations` returns that cached frame when all requested columns are present and no annotation source is being refetched. An existing `ted_domains` value such as `unclassified|90.5` therefore bypasses the updated `TedRetriever._format_domains`. `--refetch ted` is the correct targeted invalidation path because it drops the cached TED columns before retrieval. Required direction: add upgrade/cache guidance—at minimum to the OpenSpec migration plan, and preferably user-facing annotation/CLI documentation—telling users who reuse an existing output directory to run `protspace prepare ... --refetch ted`. No thread state changed.

Implemented in `c98f66f3`: the OpenSpec migration plan and both user-facing annotation references now document that existing formatted TED caches retain the old value and require a one-time `protspace prepare ... --refetch ted`. The generated website reference was updated through its source. Verification: `openspec validate fix-ted-unlabeled-domain-name --strict`; `pnpm precommit`; TED retriever module (9 passed); Python non-slow suite (787 passed, 6 deselected); Ruff check and format check.

tsenoner:
## Adversarial review

Reviewed in an isolated worktree by three independent lenses (code quality, adversarial correctness, issue-resolution audit), with every finding then put through a refuter whose default position was that it is a false positive. 2 raised, 1 survived refutation.

### Applied and pushed (`a66b3474`)

- Fix 1 (verifier-revised) in /Users/tsenoner/Documents/projects/protspace-suite/.protspace-wt-pr399/apps/protspace/src/protspace/data/annotations/retrievers/ted_retriever.py: added module-level `_UNLABELED_CATH = "-"` sentinel (with comment) right after `TED_ANNOTATIONS`, and rewrote `_format_domains` to a single `parts.append(f"{label}|{plddt:.1f}")` emit site using `domain.get("cath_label") or _UNLABELED_CATH`, a conditional `name` lookup, and a `label` expression. The `if not domains: return ""` guard was KEPT per the revised instruction. Docstring updated to document all three output shapes.

The author's triage is correct and I verified it independently rather than taking it on faith.
The stated root cause ("`_format_domains` explicitly replaced a missing or `-` CATH label with the
literal `unclassified` at the backend serialization boundary") is accurate and complete — a repo-
wide grep confirms there is no second call site and no frontend translation layer that would need a
matching change.
The key design decision (design.md:24-30: fix at the retrieval boundary, reject the "translate in
the frontend" alternative) holds up under inspection. I confirmed that the same serialized string is
consumed by three separate display paths — the Dash legend via `to_display_value`, the web legend
via `parseAnnotationValue`, and CSV/bundle exports via `strip_scores_from_df` (which explicitly
lists `ted_domains`). A frontend-only substitution really would have left exports carrying
`unclassified`, so the alternative was correctly rejected.
The PR did not narrow, redirect, or decline any part of the issue — the issue is a single, narrow
request and the PR addresses it directly. No "not reproducible" claims were made, and I could not
find any part of the reported symptom that the diff leaves standing in the generator.
Two things the triage under-weights rather than gets wrong: (a) it treats the stale-cache upgrade
path as a pure documentation problem when it is the default path a returning user hits (`keep_tmp`
defaults to True), and (b) it does not account for the one committed dataset in the repo that still
carries the old literal. Neither is a misdiagnosis; both are scope choices I'd call defensible but
worth follow-ups.
On closing semantics: "Closes #342" IS warranted. The symptom has exactly one producer and that
producer is fixed, with every downstream consumer verified to pass `-` through unchanged. I would,
however, add a one-line note to the issue telling the reporter to pass `--refetch ted` when they
verify — otherwise they will re-run their original command, hit the cache short-circuit, still see
`unclassified`, and re-open this.

- **The upgrade path is documentation-only, and it is the DEFAULT path. `keep_tmp` defaults to True (apps/protspace/src/protspace/cli/prepare.py:337), and apps/protspace/src/protspace/data/processors/pipeline.py:411-436 short-circuits to `logger.warning("Using cached annotations")` whenever the cached columns already cover the request and no refetch stage was named. Nothing stamps a formatter/cache version into `{output}/tmp/all_annotations.parquet`, so the cache cannot know its `ted_domains` column was produced by the old formatter.**
  - _Why it matters:_ The person who filed #342 will upgrade protspace, re-run their exact original `protspace prepare` command against the same output directory, and still see `unclassified` — with no warning and no hint that anything is stale. From their point of view the fix did not ship. The two doc lines added by this PR are the only thing standing between them and re-opening the issue, and nothing in the CLI surfaces those lines at the moment they matter.
  - _Suggested follow-up:_ Stamp a per-source formatter/schema version into the annotation cache parquet (analogous to `stamp_format_version` / `FORMAT_VERSION_KEY` in data/annotations/encoding.py) and auto-invalidate a source's cached columns when its stamped version is older than the current one. Minimum viable alternative: when the cache short-circuit fires and a cached `ted_domains` cell still contains the literal `unclassified`, log a one-line hint pointing at `--refetch ted`.
- **A committed, deployed dataset still contains the old literal. `apps/web/public/data/phosphatase.parquetbundle` has 407 of 1587 rows whose `ted_domains` value still reads e.g. `2.30.29.30 (Pleckstrin-homology domain (PH domain)/Phosphotyrosine-binding domain (PTB))|91.8;unclassified|93.0`. The file is tracked in git, `apps/web/vite.config.ts` sets no publicDir exclusion, so it is copied into the build and served at `/data/phosphatase.parquetbundle`.**
  - _Why it matters:_ It is the only artifact left in the repo that still exhibits the exact symptom the issue reports, and it is publicly reachable. Severity is genuinely low — it is wired only to the WebGL perf suite (apps/web/src/perf/webgl-perf-suite.ts:75-90 and apps/web/public/data/datasets.json), not to the app's dataset picker, and the app's default `apps/web/public/data.parquetbundle` has no `ted_domains` column at all — but anyone who loads that bundle by URL or drag-and-drop sees `unclassified` in the legend on a supposedly fixed build.
  - _Suggested follow-up:_ Either regenerate `apps/web/public/data/phosphatase.parquetbundle` with the fixed formatter, or (cheaper, given the binary churn) note in the PR that the perf-only fixtures are intentionally frozen and are not user-facing example data.
- **The spec scenario that matches the actual reported protein has no test. openspec/changes/fix-ted-unlabeled-domain-name/specs/ted-domain-annotations/spec.md:14-17 declares "An unlabeled domain appears with labeled domains" — which is precisely the W6JQJ9 shape (labeled, unlabeled, labeled) — but no test covers it. `test_multiple_domains` uses two labeled domains; `test_unlabeled_domain_preserves_ted_label` uses a single unlabeled domain in isolation. tasks.md 1.1 nonetheless claims the regression coverage is in place.**
  - _Why it matters:_ The mixed cell is where a regression would actually bite: a `-` hit sitting between two `;`-joined labeled hits is the case that interacts with the paren-aware splitter (conversion.ts:501-532 `splitOnTopLevelSemicolons`) and with the CATH names that legitimately contain `(` and `;`. The two existing tests would both still pass if the joining or splitting of a mixed cell broke. A declared spec scenario with no executable counterpart is also an OpenSpec accuracy problem, independent of the bug.
  - _Suggested follow-up:_ Add a test to apps/protspace/tests/test_ted_retriever.py using the real W6JQJ9 domain shape — `[_make_domain('3.40.50.2000', 94.1909), _make_domain('-', 96.7064), _make_domain('3.40.50.2000', 95.113)]` — asserting the exact serialized string, so spec scenario 2 is actually locked in.
- **No frontend test pins that `-` survives as a legend label. The claim holds today only because `-` is absent from `MISSING_VALUE_TOKENS` in packages/utils/src/visualization/missing-values.ts:13-19 — a set whose whole stated purpose (per its own header comment) is that contributors add spellings to it.**
  - _Why it matters:_ `-` is an unusually plausible candidate for someone to add to a missing-value token set in a future PR; it reads like a null placeholder. If that happens, `normalizeMissingValue` at conversion.ts:562 starts filtering the hit and TED's unlabeled domains silently revert to disappearing (or folding into N/A), quietly regressing #342 with no test failing. The existing conversion-numeric.test.ts:348 case guards `-` for numeric columns only, not for a piped categorical hit.
  - _Suggested follow-up:_ Add one case to packages/core/src/components/data-loader/utils/conversion.test.ts feeding `ted_domains: '3.40.50.2000 (P-loop NTPases)|94.2;-|96.7'` and asserting `result.annotations.ted_domains.values` contains both the labeled value and `-`, with `-` distinct from NA_VALUE.
- **The value change is user-visible but carries no migration for persisted styling and no release note beyond the `fix:` subject. Legend colors/shapes are persisted keyed by category value (bundle settings table via data/io/settings_converter.py, and the web legend's persistence-controller), so a user who recolored the `unclassified` category now has an orphaned key and gets a default color for `-`. Separately, `unclassified` was self-describing in a legend; a bare `-` is not.**
  - _Why it matters:_ This is exactly the risk design.md:38-40 identifies ("Existing consumers may group on the literal `unclassified` value") and then discharges with documentation alone. The silent color reset is minor but confusing on a re-generated dataset, and a lone `-` swatch in a legend gives the user no way to know it means "no CATH superfamily assigned" unless they follow the docs link on the annotation.
  - _Suggested follow-up:_ Confirm this is an accepted trade-off (it likely is, since the issue explicitly requests the bare `-`), and consider surfacing the meaning where the user actually is — e.g. a tooltip/aria description for the `-` TED category, reusing the wording already added to packages/utils/src/visualization/annotation-metadata.ts's docsUrl target.

</details>

### Findings needing a decision

None — everything that survived refutation was a safe cleanup and has been applied.

_1 further finding(s) were raised and refuted during verification._

FlorinSenoner:
Disposition for the newest adversarial-review round (comment 5203112758), implemented where justified in `31c5a934`:

- **Already-pushed formatter refactor:** retained as-is. The `_UNLABELED_CATH` sentinel, single emit site, empty-domain guard, and documented fallback shapes at `a66b3474` remain intact.
- **Legacy cache guidance — actionable:** the cache short-circuit now detects the precise old domain-boundary `unclassified|score` shape and warns users to run `--refetch ted`. It does not mutate or broadly version unrelated cached annotations. A regression test exercises the real parquet cache path.
- **Committed phosphatase bundle — actionable:** refreshed the public/downloadable bundle in place. 407 rows / 468 TED domain hits changed from `unclassified` to `-`; all 40 non-TED columns, schema metadata, and both projection parquet parts remain unchanged. The TypeScript reader smoke test loaded all 1,587 proteins with `-` present and no `unclassified` category.
- **Mixed labeled/unlabeled spec scenario — actionable:** added the W6JQJ9-shaped backend regression, asserting exact source order and serialization: `3.40.50.2000|94.2;-|96.7;3.40.50.2000|95.1`.
- **Frontend missing-token claim — technically unsound as stated:** I reproduced the suggested future mutation by adding `-` to `MISSING_VALUE_TOKENS`. The proposed `-|96.7` conversion test still passed because missing-value filtering sees the complete hit `-|96.7`, not the parsed label `-`. I removed that immediately-green test rather than commit a speculative change detector.
- **Persisted styling / tooltip — accepted scope trade-off, no code change:** issue #342 explicitly requires the category value to become bare `-`, and this change's OpenSpec excludes frontend rendering changes. Automatically remapping arbitrary persisted `unclassified` keys or adding TED-specific legend UI would introduce broader migration/rendering behavior without a source-safe rule.
- **Issue #342 follow-up note:** not posted because this task is restricted to PR #399. The new runtime warning now supplies the `--refetch ted` instruction at the point a returning user actually hits the stale cache.

Verification: both new regressions were observed red for their intended missing behaviors and green after the fix; focused Python suite `89 passed`; full non-slow Python workspace suite `789 passed, 6 deselected`; Ruff lint/format clean; strict OpenSpec validation clean; `pnpm precommit` clean (and reran successfully in the commit hook). No review-thread resolution state was changed.

PR #400, opened 2026-08-01
fix(annotations): preserve missing pdb availability

### LINKED ISSUE #337: [BUG] Has PDB structure annotation is True for proteins without UniProt ID
### Bug Description

Starting from a dataset with sequences mapped to UniProt and sequences not mapped to UniProt (custom naming) the "Has PDB structure" annotation is still True for every data point. As this information is (probably) derived from UniProt it should be N/A for proteins not mapped to UniProt IDs.

### Code of Conduct

FlorinSenoner:
## Summary

"Has PDB structure" reported a definite value for proteins that have no UniProt mapping at all. PDB availability is derived entirely from UniProt, so a protein with a custom identifier has no evidence either way — but the pipeline emitted a hard `False`, turning absence of evidence into evidence of absence.

`xref_pdb` now carries three states, and the persisted annotation cache learned how to migrate values written under the old contract.

Closes #337

## Root cause

`transform_xref_pdb()` received only the raw PDB cross-reference value. Two genuinely different situations arrived as the same empty string:

| situation | raw `xref_pdb` | before | after |
| --- | --- | --- | --- |
| no UniProt entry at all | `""` | `False` | `""` → rendered `N/A` |
| resolved entry, no PDB xref | `""` | `False` | `False` |
| resolved entry, has PDB xref | `"1A7F;…"` | `True` | `True` |

The transform now takes the sibling `uniprot_kb_id` as resolution context: an explicitly blank identifier means the protein has no UniProt entry, so PDB availability stays missing. The frontend's existing missing-value path already renders that as `N/A`, so no wire-format or rendering change was needed. Mapped entries are untouched.

## What changed

**The fix itself**

- Three-state `transform_xref_pdb`, keyed on the resolved UniProt identifier.
- Transforms are idempotent for canonical `"True"`/`"False"`. Cached annotations are read back and re-transformed on resumed runs, and a stored `"False"` is a non-empty string that would otherwise flip to `"True"` on the second pass. `signal_peptide` had the same exposure and got the same treatment.

**Cache migration.** Values already written by the old code cannot be repaired locally — a stored `"True"` may be a genuine PDB hit *or* a double-transformed `"False"`, with no local discriminator. So `all_annotations.parquet` carries a semantics version in its Parquet metadata:

- `CACHE_SEMANTICS_CHANGES` maps each version to the columns whose stored meaning changed at it, and `ANNOTATION_CACHE_VERSION` is derived from its highest key, so the marker and its meaning cannot drift apart. The pipeline derives the stale columns and the sources to refresh from that table alone — nothing names `xref_pdb`.
- The refresh is scoped to runs that actually surface the affected column. `xref_pdb` is absent from the `default` annotation group, so an unscoped migration would have made most runs against an `-a all` cache pay a full UniProt refetch for a column nobody reads. When a stale column is unused it is dropped from the cached frame instead, so it cannot ride along into a cache this run stamps as current; a later run that requests it sees it as missing and fetches it normally.
- A failed refresh preserves the existing cache *and* falls back to its values for the current run, so a transient outage cannot leave a run worse off than the cache it declined to overwrite. `xref_pdb` is excluded from that fallback — repairing its ambiguity is the whole point — so it stays unknown until a refresh succeeds. An explicit `--refetch uniprot` is not second-guessed.

**Two defects found during review**

- `DataFormatter.to_dataframe` derived its schema from `proteins[0]`, which is the schema path for `protspace annotate`, `prepare` without `--keep-tmp`, and the suppressed migration-failure branch. Unresolved identifiers are appended first and carry no taxonomy keys, so `-a taxonomy` on a set containing one non-accession header silently dropped `genus`/`family` from the entire output. Schema derivation now lives in `DataFormatter` and `AnnotationWriter` calls it, so the two paths cannot disagree.
- `transform_xref_pdb` compared the mapping context with `== ""` and tested the value for truthiness. Both fail for a pandas `NaN` — truthy, and stringifies to `"nan"` — so a null cell reported "Has PDB structure = True". Blankness is now tested explicitly, with `None` still meaning "no context supplied".

**Quality pass.** Consolidated the cache-version constants beside the existing `BUNDLE_FORMAT_VERSION`; named the persisted `"True"`/`"False"` vocabulary once and shared it; removed a write-only failure counter and a production-side guard that existed only to tolerate bare `MagicMock`s; shared header/row construction between `write_csv` and `write_parquet`; flattened the refetch/migration branch; finished the half-applied `TAXONOMY_LOOKUP_ANNOTATION` rename.

## Validation

Verified end to end on a 149-sequence glycosyltransferase set mixing 37 UniProt accessions with 112 custom identifiers — the exact shape #337 describes:

```
xref_pdb       N/A   False   True
resolved         0      33      4
unresolved     112       0      0
```

Against a reconstructed pre-fix cache (145 `False` + 4 `True`, of which 112 were fabricated):

- `-a xref_pdb` → migration runs, all 112 fabricated `False` repaired to `N/A`, cache stamped v1
- `-a gene_name` → no UniProt refetch, column dropped, cache left unversioned so a later run still migrates
- rerun on the migrated cache → fast path, no refetch, no migration loop
- simulated outage mid-migration → all 22 cached `gene_name` values survive, `xref_pdb` honestly `N/A`, cache left for retry
- `-a all` rebuild → incremental path fetched only the 24 missing annotations with no re-migration of the v1-stamped UniProt columns

Every regression test added here was confirmed red on the pre-fix code and green after.

- Python suite: `793 passed, 2 skipped`
- Ruff check + format, Prettier, `pnpm precommit` (typecheck, Knip, generated annotation check, docs build)
- `openspec validate --all --strict`: 14 passed, 0 failed

## Merging

⚠️ **Do not squash.** This PR touches `apps/protspace/` across 14 commits; squashing collapses them into one commit touching every path, which destroys semantic-release's per-commit path scoping and cuts a spurious PyPI release (observed on #387 → v4.9.1). Use a merge commit or rebase merge — see AGENTS.md.


apps/protspace/src/protspace/data/annotations/transformers/transformer.py
```diff
@@ -76,9 +76,12 @@ def _transform_annotations(self, annotations: dict) -> dict:
             )
 
         if "xref_pdb" in transformed:
-            transformed["xref_pdb"] = self.uniprot_transformer.transform_xref_pdb(
-                transformed["xref_pdb"]
-            )
+            if transformed.get("uniprot_kb_id") == "":
+                transformed["xref_pdb"] = ""
+            else:
+                transformed["xref_pdb"] = self.uniprot_transformer.transform_xref_pdb(
```
FlorinSenoner: [P1] Keep cached PDB values idempotent

`ProteinAnnotationManager` rehydrates `all_annotations.parquet` through `_extract_cached_source()` whenever another source is missing or refetched, then runs this transformer again. Those persisted `xref_pdb` cells are already canonical `""`, `"False"`, and `"True"`; this branch feeds `"False"` back to `transform_xref_pdb()`, whose truthiness check returns `"True"`. I reproduced `to_pd()` returning `["", "True", "True"]` from those three cached inputs, so a resolved protein with no PDB becomes a false positive (and this can recreate the reported all-`True` plot) even though the new direct unit test passes. Please make the transform idempotent for canonical `True`/`False` values or skip raw transforms for cached UniProt data, and add a manager/cache round-trip test covering all three states.

FlorinSenoner: Fixed in ee02c868. `transform_xref_pdb()` now preserves persisted canonical `"False"` and `"True"` values before applying raw PDB-ID truthiness conversion. Added a real parquet-backed `ProteinAnnotationManager` cache round-trip covering `["", "False", "True"]`; it failed as `["", "True", "True"]` before the fix and now passes. Fresh local evidence: affected slice 164 passed, full suite 791 passed / 6 slow deselected, Ruff clean across 143 files, strict OpenSpec valid, `pnpm precommit` green, and bundle contract 11 passed.

tsenoner:
### Automated review

**Does it solve #337?** Partially — the PR fixes both mechanisms behind the all-`True` symptom: `transformer.py` now emits an empty `xref_pdb` when the sibling `uniprot_kb_id` is empty, and `transform_xref_pdb` returns canonical `"False"`/`"True"` unchanged, which stops the second transform pass over cached annotations from turning a stored `"False"` into `"True"` (I confirmed the reporter's bundle has 112 rows with a fully empty UniProt row yet `xref_pdb == "True"`, only reachable as `"" → "False" → "True"`). Any fresh run is fully fixed. What it misses is the data already on disk: the wrong values live in `<output>/tmp/all_annotations.parquet`, and re-running the same `prepare` hits the cache fast path, so the reporter still sees all-`True` unless they delete `tmp/` or pass `--refetch uniprot`.

Found 3 issues:

1. Annotation caches already containing the wrong values are never invalidated, so an identical re-run still emits the reported all-`True` column — `prepare` defaults to `keep_tmp=True`, and when nothing is missing and no `--refetch` is given `_fetch_annotations` returns `cached_df[cols]` without running the transformer at all. On a *partial* re-run the new canonical guard additionally freezes a wrongly-flipped `"True"` in place for mapped proteins, so design.md's "No data migration is required" is not accurate for the user who filed the issue.

https://github.com/tsenoner/protspace/blob/ee02c86864ae3b6d3568a9b29e05779bfeef5b97/apps/protspace/src/protspace/data/processors/pipeline.py#L411-L437

2. `transform_signal_peptide` has the exact non-idempotency this PR just fixed for `xref_pdb`, but in the damaging direction: cached InterPro columns are re-transformed whenever `sources_to_fetch["interpro"]` is False, and the stored `"True"` does not contain `"SIGNAL_PEPTIDE"`, so every positive silently becomes `"False"` and is written back to `all_annotations.parquet`. This is reachable by the very workaround for #337 (`--refetch uniprot` keeps `interpro=False`). The same canonical-value guard would close the class rather than one member.

https://github.com/tsenoner/protspace/blob/ee02c86864ae3b6d3568a9b29e05779bfeef5b97/apps/protspace/src/protspace/data/annotations/transformers/interpro_transforms.py#L99-L117

3. User-facing docs still describe `xref_pdb` as a strict two-valued boolean after the domain gained an empty state — same wording in `docs/scripts/annotation-details.ts` L115-L119, which generates `docs/guide/annotations.md` L179 and surfaces in the app. `pnpm docs:annotations:check` only checks generated-vs-source, so green precommit says nothing about accuracy.

https://github.com/tsenoner/protspace/blob/ee02c86864ae3b6d3568a9b29e05779bfeef5b97/apps/protspace/docs/annotations.md#L81-L96

<sub>Reviewed at ee02c86864ae3b6d3568a9b29e05779bfeef5b97 against issue #337.</sub>

FlorinSenoner:
Reviewed against current head `ee02c86864ae3b6d3568a9b29e05779bfeef5b97`. All three items are actionable:

1. Confirmed: `_fetch_annotations()` returns a complete `all_annotations.parquet` cache before transformation, so an existing bad `xref_pdb` cache survives an identical rerun. On mixed reruns, unmapped rows can be corrected from `uniprot_kb_id`, but a previously corrupted mapped `"True"` is indistinguishable from a legitimate positive. The required direction is to invalidate/refetch or explicitly migrate affected UniProt caches and correct the OpenSpec migration plan.
2. Confirmed: the current `transform_signal_peptide()` maps raw `"SIGNAL_PEPTIDE"` to `"True"` but maps persisted `"True"` to `"False"`. `--refetch uniprot` retains cached InterPro values and then runs the shared transformer, so this path can overwrite positives. The required direction is an idempotent canonical-value path (or raw/cached separation) plus a mixed-cache/refetch regression.
3. Confirmed: the OpenSpec requirement defines empty/N/A for unresolved entries, while `apps/protspace/docs/annotations.md` and `docs/scripts/annotation-details.ts` still document only `True`/`False`. The required direction is to update both canonical documentation sources and regenerate the derived guide.

No repository or PR state changes were made beyond this requested triage response.

FlorinSenoner:
Implemented the three verified follow-ups in 3fbe03a8:

1. Legacy all_annotations.parquet files containing xref_pdb now bypass the cache fast path, refresh only UniProt columns, preserve other cached sources, and are rewritten with protspace_annotation_cache_version = 1 so later runs reuse them normally.
2. Cached signal_peptide True/False values are now idempotent during mixed cache/refetch runs, with manager-level regression coverage.
3. OpenSpec migration/cache semantics and all canonical/generated xref_pdb docs now describe the empty/N/A state.

Verification: focused annotation/pipeline slice 220 passed; Python CI-equivalent suite 793 passed / 6 slow deselected; Ruff check and format-check passed for 143 files; bundle contract 11 passed; strict OpenSpec valid; pnpm precommit passed including lint, type-check, Knip, generated-doc check, and docs build.

apps/protspace/src/protspace/data/processors/pipeline.py
```diff
@@ -443,11 +452,18 @@ def _fetch_annotations(
                     cached_annotations, required
                 )
 
-                if refetching_annotations:
+                if refetching_annotations or legacy_pdb_cache:
                     # Override with explicitly requested sources
                     sources = {src: src in refetch for src in _ANN_SOURCES}
+                    if legacy_pdb_cache:
```
FlorinSenoner: [P1] Preserve cached taxonomy during the UniProt migration

This migration later drops `organism_id` because that join key is UniProt-owned. When taxonomy is not explicitly refetched, `ProteinAnnotationManager` then calls `_extract_cached_taxonomy()` with a frame that no longer has `organism_id`, gets `{}`, and silently omits the cached taxonomy columns from both this result and the rewritten cache. I reproduced a legacy cache containing `xref_pdb`, `organism_id=9606`, `genus=Homo`, and cached `signal_peptide`: after `_fetch_annotations()`, `signal_peptide` survived and the cache was stamped version 1, but `genus` disappeared from the returned and persisted columns. Please retain/re-hydrate the taxonomy join context while refreshing UniProt and add a migration regression covering a taxonomy column; the change spec explicitly requires every unaffected source to be preserved.

FlorinSenoner: Fixed in f42426e8. During a UniProt-only cache migration, source-scoped cleanup now retains the shared organism_id lookup key only when cached taxonomy is being reused. The manager rehydrates cached taxonomy before the refreshed merge; cached UniProt records remain excluded, and the freshly fetched organism_id selects the taxonomy entry. The production-path Parquet regression preserves genus=Homo and cached signal-peptide values, replaces a stale UniProt gene, stamps version 1, and verifies the second run takes the cache fast path. Verification: focused 1 passed; affected 220 passed; full non-slow 793 passed / 6 deselected; Ruff clean across 143 files; bundle contract 11 passed; strict OpenSpec valid; generated docs/docs build and pnpm precommit passed.

apps/protspace/tests/test_pipeline_utils.py
```diff
@@ -345,6 +346,88 @@ def test_tsv_path(self):
...
+            PipelineConfig(
+                methods=[],
+                output_path=tmp_path / "unused.parquetbundle",
+                annotations=["xref_pdb", "genus", "signal_peptide"],
+                keep_tmp=True,
+                intermediate_dir=tmp_path,
+            )
+        )
+
+        headers = ["resolved_without_pdb", "unresolved"]
```
FlorinSenoner: [P1] Cover the unresolved-first ordering before stamping this cache migrated. AnnotationMerger and AnnotationTransformer preserve row order and non-uniform annotation keys, while AnnotationWriter.write_parquet() derives the Parquet columns from proteins[0].annotations.keys(). With these rows reversed, the unresolved first protein has no cached taxonomy keys, so the later resolved protein’s preserved genus is omitted from the rewritten cache even though the migration is marked current. Please make schema construction order-independent and add the reversed mixed mapped/unmapped regression.

FlorinSenoner: Fixed in 2b2c88e3a575dbffd9a70bf64eb2cb4904da0501. AnnotationWriter now builds a first-seen ordered union of annotation keys across every protein, keeping serialization generic while preventing later-only taxonomy fields from being dropped. The production-path migration regression now runs unresolved-first and verifies genus survives both the rewritten cache and subsequent fast-path reuse. Verification: focused and affected tests passed; full Python 793 passed with 6 slow deselected; Ruff, bundle contract (11), strict OpenSpec, docs build, and pnpm precommit passed. Exact-head GitHub checks are terminal: 8 passed, 2 intentionally skipped.

tsenoner:
## Adversarial review

Reviewed in an isolated worktree by three independent lenses (code quality, adversarial correctness, issue-resolution audit), with every finding then put through a refuter whose default position was that it is a false positive. 10 raised, 6 survived refutation.

### Applied and pushed (`b885e6f1`)

- Fix 1 (pipeline.py): merged the standalone `from protspace.data.annotations.configuration import (TAXONOMY_LOOKUP_ANNOTATION,)` statement (was lines 469-471) into the existing configuration import at lines 447-450, which unconditionally dominates the usage site. Left the `AnnotationConfiguration as AnnCfg` aliased import and the `AnnCfg.categorize_annotations_by_source` call untouched per the authoritative revision. Three import statements became two; no name changes.
- Fix 2 (annotations/manager.py line 189): `internal_columns = ["organism_id", "sequence"]` -> `internal_columns = [TAXONOMY_LOOKUP_ANNOTATION, "sequence"]` (constant already imported at line 13).
- Fix 2 (annotations/manager.py line 221): `ProteinAnnotations(identifier=header, annotations={"organism_id": ""})` -> uses `TAXONOMY_LOOKUP_ANNOTATION` as the key. Substituting the longer constant pushed the line past the 88-char limit, so it was wrapped across three lines exactly as `ruff format` produces (verified by `ruff format --check` reporting the file already formatted). merging.py:75 was left alone and the constant was not renamed, per the authoritative revision.

The scoping is correct and the fix is aimed at the right layer (producer semantics, not a frontend
coercion). Restricting the change to `xref_pdb` is justified: I checked the other UniProt-derived
fields and none fabricates a negative for an unmapped protein — `transform_fragment("")`,
`transform_annotation_score("")`, `transform_cc_subcellular_location("")` and the GO/keyword/ec
transforms all return empty, and `signal_peptide` only sees a value when InterPro actually returned
a record for that identifier (`_merge_interpro` adds keys only for identifiers present in the
results dict). `xref_pdb` genuinely was the only field asserting a UniProt fact without UniProt
evidence.
Where the triage is wrong is the "not reproducible" caveat. The PR text still says the reported
wording ("True for every data point") "was not produced by this minimal current-main dataset". That
is stale and misleading: I reproduced exactly that symptom on the production code path (cached
"False" -> "True" for every mapped protein whenever a run reuses the UniProt cache while fetching
another source). The author evidently found it later — commits ee02c868/3fbe03a8 and the design.md
section "Make PDB canonicalization idempotent across cache reads" exist precisely for it — but the
description was never updated. It should be, because the reporter's symptom needs a cache-reuse run
to appear, and that is also exactly the population the new cache migration protects. A reviewer
reading the description would wrongly conclude the reporter's observation remained unexplained.
"Closes #337" is warranted. Both halves of the reported behaviour are addressed for newly produced
data, and the stale-cache trap that would otherwise re-inject the bug after upgrade is handled
automatically.

- **The legacy-cache migration throws away incremental source selection and silently drops newly requested annotations from other sources. pipeline.py:455-459 replaces the computed `sources` with `{src: src in refetch}` (empty when no --refetch was passed) and only forces `uniprot=True`. Verified by running `ReductionPipeline._fetch_annotations` against an unstamped cache holding UniProt columns while requesting `xref_pdb,pfam`: only uniprot was fetched, InterPro was never called, and the returned frame has no `pfam` column at all (['identifier','xref_pdb','gene_name','protein_name','uniprot_kb_id']). On main the same input yields {'interpro': True} from `determine_sources_to_fetch`, so this is a regression introduced by this PR.**
  - _Why it matters:_ Every existing cache is 'legacy' after upgrade, so the first post-upgrade run that also asks for a new annotation source (pfam/cath from InterPro, organism_name from taxonomy, TED, Biocentral) silently produces a bundle missing that annotation. There is no error and no 'Missing annotations:' log — that branch is skipped. It self-heals on the next run, which makes it look like a flaky tool rather than a bug.
  - _Suggested follow-up:_ Union the migration override with the incremental result instead of replacing it: keep `sources = determine_sources_to_fetch(...)`, then OR in `src in refetch` and the forced uniprot. Add a regression test for 'legacy cache + newly requested InterPro annotation' asserting the pfam column survives.
- **Migration triggers on cache *content*, not on what the run needs: any unstamped cache containing `xref_pdb` forces a full UniProt refetch of every header even when `xref_pdb` is not requested this run (e.g. run 1 `-a uniprot`, run 2 `-a default`).**
  - _Why it matters:_ For a Swiss-Prot-scale input this is a multi-hour, whole-dataset re-hit of the UniProt REST API announced only by a single logger.warning, on a run whose output does not even contain the affected column. apps/protspace/docs/cli.md:356 presents it as a narrow one-time refresh, which undersells the cost.
  - _Suggested follow-up:_ Gate `legacy_pdb_cache` on `'xref_pdb' in required` as well as in `cached_annotations`, or drop just the `xref_pdb` column and let the existing missing-annotation logic decide; either way state the expected fetch size in the warning.
- **Only `write_parquet` got the order-independent schema fix. `write_csv` (writers.py:53) still derives headers from `proteins[0].annotations.keys()` and does not accept/propagate `dataframe_attrs`.**
  - _Why it matters:_ design.md states the writer derives the schema 'from every annotation record', but that now holds for one of two writers. `write_csv` has no caller in src today, so this is latent — yet if it is ever wired up (or used by a consumer of the public `AnnotationWriter`), a heterogeneous first record truncates columns exactly the way commit 2b2c88e3 just fixed for parquet.
  - _Suggested follow-up:_ Extract the header-union computation into one helper shared by both writers, or delete `write_csv` if it is genuinely dead.
- **Already-generated .parquetbundle files are not migrated and carry no signal that their `xref_pdb` column predates the three-state contract.**
  - _Why it matters:_ The reporter of #337 already has a bundle where unmapped proteins say False (or, on a cache-reuse run, where everything says True). Nothing in the app tells them the column is wrong; they must know to re-run prepare. The bundle format version (encoding.py BUNDLE_FORMAT_VERSION = 2) was not bumped and the diff contains no release note.
  - _Suggested follow-up:_ Call it out in the next protspace release notes ('re-run prepare to correct xref_pdb for unmapped proteins'), and decide whether the frontend should flag a bundle whose xref_pdb column has zero empty cells — or record it as accepted and out of scope.
- **The cache-version marker rides on `DataFrame.attrs`, which is preserved by pyarrow's pandas metadata (pandas_compat.py:280) rather than by parquet itself. apps/protspace/pyproject.toml pins pyarrow>=20.0.0 (fine today) but also pandas>=2.0.0, and any rewrite of all_annotations.parquet by a tool that does not carry pandas metadata drops the marker.**
  - _Why it matters:_ A dropped marker is indistinguishable from a legacy cache, so the whole-dataset UniProt refetch repeats on every run — an effectively permanent loss of the annotation cache with only a warning line to show for it. design.md acknowledges the case but treats 're-run the safe refresh' as free, which it is not at 570k proteins.
  - _Suggested follow-up:_ Write the marker into the parquet file's own key-value metadata via pyarrow (the pattern already used by `stamp_format_version` in encoding.py) instead of `df.attrs`, or at minimum add a test pinning the round-trip so a pyarrow bump cannot silently break it.

</details>

### Findings needing a decision (4)

#### 1. The legacy-cache migration piggybacks on the `--refetch` override branch, discarding the `determine_sources_to_fetch` result and silently skipping genuinely missing non-UniProt sources.

`apps/protspace/src/protspace/data/processors/pipeline.py:455` · **high** · altitude

`sources = AnnotationConfiguration.determine_sources_to_fetch(cached_annotations, required)` (line
451) computes which sources must be fetched because their annotations are missing from the cache.
The new condition `if refetching_annotations or legacy_pdb_cache:` then *replaces* that dict
wholesale with `{src: src in refetch ...}`. For the migration-only path `refetch` is empty, so every
source becomes False and only `uniprot` is set back to True. Concrete failure: user has a pre-
version cache from `--annotations uniprot` (contains `xref_pdb`, no `genus`), then runs `prepare
--annotations uniprot,taxonomy`. `missing = {genus, ...}` so `determine_sources_to_fetch` returns
`taxonomy=True`, but the override zeroes it; only UniProt is fetched, `genus` is never retrieved,
and the final `annotations_to_keep` filter in `ProteinAnnotationManager.to_pd` silently drops it
from the returned frame and the bundle — with no warning, because the `logger.info(f"Missing
annotations: {missing}")` line only runs in the `else` branch. The run self-heals only on a second
invocation (the cache is now stamped). Root cause is altitude: a one-time migration flag was bolted
onto a branch whose semantics are "user explicitly asked to replace the source set", instead of
being expressed as an additional requirement on top of the derived source set.

In apps/protspace/src/protspace/data/processors/pipeline.py, make the migration additive instead of
an override. Replace the body of the `if refetching_annotations or legacy_pdb_cache:` branch's first
lines with:
                if refetching_annotations or legacy_pdb_cache:
                    if refetching_annotations:
                        # Override with explicitly requested sources
                        sources = {src: src in refetch for src in _ANN_SOURCES}
                    if legacy_pdb_cache:
                        sources["uniprot"] = True
                        logger.warning(
                            "Refreshing legacy UniProt annotation cache to apply "
                            "current PDB availability semantics"
                        )
                    refetched = [s for s in _ANN_SOURCES if sources[s]]
                    if refetching_annotations:
                        logger.info(f"--refetch: re-fetching {', '.join(refetched)}")
and leave the rest of the branch (the cols_to_drop loop and the taxonomy discard) unchanged. Add a
regression test: legacy cache (no version attr) containing xref_pdb but not `genus`,
annotations=["xref_pdb", "genus"] -> TaxonomyRetriever is called and the returned frame has a
`genus` column. Not auto-applicable: it changes fetch behavior and needs the new test.

</details>

#### 2. The legacy-PDB cache migration replaces the computed `sources` map with an all-False/uniprot-only override, so annotations that are genuinely missing from a legacy cache and owned by taxonomy/InterPro/TED/Biocentral are never fetched and silently vanish from that run's output.

`apps/protspace/src/protspace/data/processors/pipeline.py:457` · **high** · correctness

`sources = AnnotationConfiguration.determine_sources_to_fetch(cached_annotations, required)` (line
450) is discarded whenever `legacy_pdb_cache` is true, not just when the user passed `--refetch`.
With an empty `refetch` set, `{src: src in refetch ...}` is all-False and only `uniprot` is re-
enabled, so every other source is turned OFF even when `missing` says it is needed. Reproduced
against the worktree: legacy cache (no version attr) with columns
identifier/xref_pdb/gene_name/protein_name/uniprot_kb_id/organism_id, running
`pipeline._fetch_annotations(["P1","P2"])` with `annotations=["xref_pdb", "genus"]` -> `missing ==
{"genus"}`, TaxonomyRetriever.fetch_annotations is called 0 times and the returned frame is
`['identifier','xref_pdb','gene_name','protein_name','uniprot_kb_id']` — `genus` is absent from the
bundle with no warning (the `logger.info(f"Missing annotations: {missing}")` branch is skipped too).
The identical run with the cache stamped `protspace_annotation_cache_version=1` fetches taxonomy and
returns `genus`, so this is a regression introduced by wiring `legacy_pdb_cache` into the
`--refetch` override. It self-heals only on a *second* prepare run (the migrated cache is now
versioned), so the first upgrade run after `-a uniprot` -> `-a uniprot,taxonomy` (or `-a all` under
a release that added new group members) silently ships a bundle missing the requested annotation. No
test covers legacy-cache + missing-annotation-from-another-source.

Same single edit as finding [0]: in apps/protspace/src/protspace/data/processors/pipeline.py, only
`refetching_annotations` may replace `sources`; `legacy_pdb_cache` may only set `sources["uniprot"]
= True` on top of the derived map. Guard the `--refetch:` log line with `if refetching_annotations:`
so the migration-only path stays quiet, and keep the existing cols_to_drop / taxonomy-discard logic
unchanged. Add the regression test described in [0]. Not auto-applicable (behavior change plus new
test).

</details>

#### 3. Cross-field PDB policy is inlined as an if/else in the orchestrator instead of being passed to `UniProtTransformer.transform_xref_pdb`, and the `== ""` test misses an absent `uniprot_kb_id` key, contradicting design.md.

`apps/protspace/src/protspace/data/annotations/transformers/transformer.py:78` · **medium** · altitude

Every other branch in `_transform_annotations` is a uniform one-line delegation to a
`*_transformer.transform_*` call, and the codebase already has an established pattern for a
transform that needs a sibling field: `transform_pfam_clan(transformed.get("pfam", ""),
self._pfam_clan_map)` on line 133. The new `xref_pdb` branch instead nests policy ("unresolved
UniProt entry means unavailable") in the dispatcher, so the rule is invisible to anyone reading
`uniprot_transforms.py` and untestable through the transformer's own unit surface. Second,
`transformed.get("uniprot_kb_id") == ""` returns False when the key is absent (`.get` yields
`None`), so a record lacking the column falls through to the truthiness conversion and is asserted
as `False`. design.md's Risks section explicitly promises the opposite: "A malformed resolved record
could omit `uniprot_kb_id` → Treat it as unavailable rather than asserting a potentially false
negative." Code and design doc disagree on a documented risk mitigation.

Pick one side and make it explicit. Cheapest correct option: amend openspec/changes/fix-pdb-
annotation-without-uniprot/design.md's Risks bullet to state that an *empty* `uniprot_kb_id` marks
PDB availability unavailable, and that a completely absent key means the UniProt context was not
supplied (partial-header transform_row) and keeps the truthiness conversion. If instead you want the
conservative behavior the doc currently promises, change transformer.py:79 to `if not
transformed.get("uniprot_kb_id"):` AND update the five tests listed above (test_transformer.py:105,
211, 279, 318, 483) to include a `uniprot_kb_id` value. Do NOT add a defaulted `uniprot_kb_id=""`
parameter to `UniProtTransformer.transform_xref_pdb` — that silently returns "" for every existing
single-arg call site. Either way requires test edits, so not auto-applicable.

</details>

#### 4. The forced migration drops and refetches all cached UniProt columns, and a partial UniProt fetch failure writes empty values that are then stamped as current-version cache, so they are served from the fast path forever.

`apps/protspace/src/protspace/data/processors/pipeline.py:459` · **medium** · edge-case

`UniProtRetriever.fetch_annotations` swallows per-batch errors: on any exception it appends
`dict.fromkeys(UNIPROT_ANNOTATIONS, "")` for the whole 100-accession batch, logs only at
`logger.debug`, and `total_batch_failures`/`total_failed_proteins` are accumulated but never
reported. Concrete scenario: a user upgrades with an existing 570K-protein
`tmp/all_annotations.parquet` written by an older release (no version attr, contains `xref_pdb`).
The migration drops every cached UniProt column (cols_to_drop) and refetches. One transient HTTP 500
on a single batch -> those 100 proteins get `gene_name`, `protein_name`, `ec`, `keyword`,
`reviewed`, `xref_pdb` = "" while previously-correct cached values are gone. `_save_and_load` then
writes that frame with `protspace_annotation_cache_version=1` and, because the union-of-headers
writer emits all UniProt columns, `missing` is empty on every subsequent run -> the `not missing and
not refetching and not legacy_pdb_cache` fast path returns the blanks permanently. The "All cached
annotations are empty" warning does not fire because only a subset of rows is blank. Before this PR
this destructive drop-and-refetch only happened under an explicit `--refetch uniprot`; the diff
makes it automatic and unavoidable for every existing cache containing `xref_pdb`.

Two changes, both outside the mechanical-cleanup bar (new plumbing plus tests):
1. apps/protspace/src/protspace/data/annotations/retrievers/uniprot_retriever.py — in the summary
block after the batch loop, surface the counters, e.g. `if total_batch_failures:
logger.warning(f"{total_batch_failures} UniProt batches failed; {total_failed_proteins} proteins
have empty UniProt annotations")`, and expose the count to the caller (e.g. store it on the
retriever instance as `self.failed_protein_count`).
2. Make the migration failure-aware: have ProteinAnnotationManager propagate that count, and in
pipeline._fetch_annotations write the ANNOTATION_CACHE_VERSION stamp only when a migration-triggered
UniProt refresh completed with zero batch failures — otherwise write to a temp path / leave the
legacy file in place so an ambiguous-but-populated cache is not replaced by silently blank rows.
Needs a test covering a partial-batch failure during migration.

</details>

_4 further finding(s) were raised and refuted during verification._

FlorinSenoner:
Addressed the 2026-08-06 adversarial review at https://github.com/tsenoner/protspace/pull/400#issuecomment-5203113015 in `b03b8f31`.

Disposition:

1. **Missing-source selection (decision items 1 and 2; duplicate audit gap): implemented.** The migration now adds its required UniProt refresh to the source plan returned by `determine_sources_to_fetch()`; only an explicit `--refetch` replaces that plan. A production-path regression starts from an unstamped PDB cache missing `genus`, verifies taxonomy is fetched during migration, and verifies `genus=Homo` is present in the same run.
2. **Missing `uniprot_kb_id` / transformer altitude (decision item 3): code behavior retained; design corrected.** Repository evidence shows `transform_row()` and both writer surfaces support partial headers, and existing tests intentionally transform `xref_pdb` without supplying `uniprot_kb_id`. An absent key therefore means mapping context was not supplied, not that resolution failed; only an explicitly empty identifier proves the unresolved state. The inaccurate OpenSpec risk bullet now states that distinction. Moving the branch into `UniProtTransformer` would not change this evidence rule and would add API churn without fixing a demonstrated behavior.
3. **Partial UniProt migration failure (decision item 4): implemented.** `UniProtRetriever` now surfaces failed batch/protein counts and emits a warning. The manager suppresses only migration-triggered cache writes after a UniProt batch failure, leaving the unversioned legacy file unstamped and unchanged so the next run retries. Ordinary first writes and explicit refetch behavior are unchanged.
4. **Migration even when the current projection omits `xref_pdb`: no change.** The accepted OpenSpec contract says an unversioned cache containing `xref_pdb` is not authoritative and receives a one-time refresh; deferring based on the current selected columns would contradict that migration decision and leave ambiguous values resident.
5. **`write_csv` schema union: no change.** The order-independence requirement and design decision are explicitly Parquet-cache-specific, and `write_csv` has no production caller under `src/`. Expanding a latent, unused surface is outside #337.
6. **Existing `.parquetbundle` migration/frontend signalling: no change.** The design names bundle-format and frontend changes as non-goals and states existing bundles are not rewritten; users correct the value by rerunning `prepare`, whose annotation cache is the migration target.
7. **Native Parquet metadata instead of pandas attrs: no change.** The current round-trip is regression-covered and the design explicitly treats missing metadata after external rewriting as legacy, causing another safe refresh. A native-metadata hardening change is separable from #337.
8. **Stale reproduction caveat in the PR body: confirmed as incomplete context, but no PR metadata was changed.** The later cache round-trip regression and commit history demonstrate the exact all-`True` path; the original sentence remains narrowly true about the stated minimal fresh-run dataset.

TDD/verification evidence:
- red: missing-source regression failed with absent `genus`; failure-safety regression found the cache incorrectly stamped version 1
- green: focused regressions `2 passed`
- affected annotation/pipeline/retriever slice: `246 passed`
- full Python non-slow suite: `795 passed, 6 deselected`
- Ruff check and format check: clean across 143 files
- strict OpenSpec validation: valid
- bundle contract: `11 passed`
- repository `pnpm precommit`: passed before commit and again in the commit hook

No review thread was resolved and no PR metadata/state was changed.

PR #402, opened 2026-08-01
feat(structure-viewer): add TED, restyle the resource row, and gate the docs-screenshot scripts

### LINKED ISSUE #293: [BUG] Annotation list inconsistency in control bar and in filter logic
### Bug Description

We have different annotation names in the control bar and in filtering logic.
e.g. in demo dataset (control bar, filter logic)
- CATH-Gene3D, cath
- SUPERFAMILY, superfamily
- EC number, ec

It's better to keep them consistent.

### Code of Conduct

### LINKED ISSUE #344: [FEATURE] Add TED link
### Feature Request

Add a link to TED besides links to UniProt and Interpro. TED links for a UniProt ID are structured like: https://ted.cathdb.info/uniprot/<UniProtID> e.g. https://ted.cathdb.info/uniprot/W6JQJ9

### Code of Conduct

FlorinSenoner:
Closes #344. Closes #293.

Started as "add a TED link". Grew, deliberately, into the three things that link exposed.

## 1. The TED link, and the row it lives in

`buildTedUrl` joins the existing builders and reuses `getBaseAccession`, so a versioned ID targets the canonical accession.

Adding a fourth link made three existing problems visible, so the row was rebuilt:

- The links rendered at `0.75rem` in `--text-secondary` — the same treatment as the accession beside them. They read as metadata, and nothing marked them as leaving the app.
- **AlphaFold wasn't in the row at all.** It was the header title, so the destination the viewer is named after was a link almost nobody would find.
- `.header-links` was a non-wrapping flex row, so each new resource made the group monolithically wider.

```
AlphaFold2 Structure   A0A023VZF1                    ✕
AlphaFold↗ · UniProt↗ · InterPro↗ · TED↗
```

The row is defined once, as `RESOURCE_LINKS` beside the URL builders, so a destination is one entry rather than an eight-line anchor plus a hand-placed `&middot;`.

**Two details only the browser caught:**

- The separator trails its item rather than leading the next, so a wrapped row starts on a label (a leading dot wrapped with the link it preceded — at a 220px sidebar the second line began `· InterPro`). It hangs off a list item, not the link: generated inside the `<a>` it became part of every link's accessible name ("AlphaFold ·"), click target and focus ring. The row is a labelled `role="list"` now; verified in Chromium's accessibility tree.
- Only `.header-link-label` is underlined. `text-decoration` propagates into descendants and can't be cancelled there, so underlining the anchor drew the line under the arrow and the separator too.

**Links are deliberately unconditional.** No availability checks, no requests. Measured first: all three probe endpoints are CORS-open and fast (`alphafold.ebi.ac.uk/api/domains` ~96ms, `rest.uniprot.org` ~281ms, `ebi.ac.uk/interpro/api` ~217ms), so feasibility wasn't the objection. Links would flicker ~250ms after each selection; a network error is indistinguishable from "no entry", so a flaky connection would hide working links; and ProtSpace never validates that an ID is a UniProt accession at all. Recorded in `design.md`.

## 2. The docs-screenshot pipeline, which was silently rotting

The TED screenshot couldn't be regenerated, because the capture suite was broken — and nothing said so.

- **The duplicate-badge capture bound to scatter-plot internals renamed in `b1695b30`.** Repaired, and `selectProjection` moved into `helpers.ts` beside `selectAnnotation`, which already owned that shape. The inline `expect.poll` it replaced compared the plot's projection to the control bar's — but `applyProjectionSelection` sets both synchronously, and PCA is `projections[0]`, so that predicate held whether or not the switch propagated.
- **`capture-helpers.spec.ts` matched no Playwright project**, so `playwright test --list` never collected it. Its assertions were thin anyway: all three gates in `waitForDataLoad` reject with the same `Timeout 100ms exceeded`, so the test claiming to pin the readiness contract passed whether or not its setup satisfied that gate. Deleted.
- **Eleven `waitForFunction` calls passed options into the `arg` slot.** The signature is `(pageFunction, arg, options)`, so every written `timeout` and `polling` value was silently ignored in favour of Playwright's defaults. Verified against `playwright-core@1.61.0`.

**Root cause: `scripts/` was in no tsconfig, matched no ESLint config, and no workflow ran it.** Now type-checked and linted via `pnpm quality` / `pnpm lint`. That immediately surfaced `headless: 'new'` (Puppeteer's string, not Playwright's boolean) and a `page` argument `logAction` never read at 20 call sites.

Hand-fixing 11 call sites wasn't enough — **a twelfth survived, in a file that same commit edited.** So there's a `no-restricted-syntax` rule instead: 1 error, 0 false positives across all 68 `waitForFunction` calls in the repo. The rule belongs to browser automation, not to `scripts/`, so `apps/web/tests` — the suite that actually runs in CI, previously matched by no ESLint glob — is covered too.

**knip was already in the quality gate and already blind:** `knip.jsonc` globbed `scripts/**/*.ts` as `entry`, making every file its own entry root. Narrowed to real roots, it found 1 unused file and 13 unused exports — more rot than both new tools combined.

## 3. A guard for something that should worry you

`content: '\00b7'` inside Lit's `css` tagged template made the cooked string `undefined` and **silently voided the entire component stylesheet**. The shadow root went from 37 CSS rules to 0. `tsc`, `knip`, all 128 core tests and the docs build stayed green, because jsdom never evaluates CSS. It was found by opening the page.

One assertion on `cssText` now covers that class for every stylesheet — verified by reintroducing the exact escape: the guard fails while all 21 structure-viewer tests still pass. The test finds sheets by glob (31 of them) rather than a hand-kept table, which had already missed three; the three components that still declared `static styles = css` inline moved into styles modules; and ESLint rejects a `css` literal anywhere else in `packages/core`, so no sheet sits out of the guard's reach.

Also fixed: `markdown.image.lazyLoading` had lazy-loaded the **LCP image** on four Explore pages, including a 3.7 MB GIF. Those four are eager with explicit dimensions; below-the-fold images still defer.

## 4. Annotation search, and labels in the query builder (#293)

Typing `ted` returned the four Biocentral predictions, because they're keyed `predic**ted**_*` and the filter matched any substring of the raw column name — text never shown on screen. The query builder had the mirror problem: it printed column names (`cath`, `ec`) where the control bar shows labels ("CATH-Gene3D", "EC number").

- **Both pickers show labels.** The query builder's picker and condition button draw an annotation through the same `renderAnnotationName` as the dropdown: the label, plus the ⚡ badge for a predicted annotation, which also tells apart the two "Subcellular location" columns. Conditions still store the column name.
- **Search matches only the displayed label**, as a case-insensitive substring, through one shared function. An intermediate version also matched column-name words; it offered `predicted` → "Membrane", "Signal peptide", … — rows with no visible "predicted" — so it was dropped. An unregistered column is labelled by its prettified name and stays findable through it.

## 5. Automation reads the scatter plot through public accessors

The docs captures and an e2e helper re-derived the plot's projection from `_plotData`/`_scales`/`_transform` in five places. One had rotted: select-single read `_plotData[idx].x` from before PlotData went columnar, got `undefined`, and clicked the plot centre every run. The plot now exposes `dataToClient` (through the interaction SVG's screen matrix — the exact inverse of the `d3.pointer` hit-test; the host rect is 1px off across its border), `getProteinClientPosition`, `getDuplicateStacks` and `getExpandedDuplicateStackKey`, and every probe uses them.

Also: one accession rule (`getBaseAccession`, now in `@protspace/utils`) for both the structure fetch and the links; `apps/web/tests` linted by the app's own `lint`; `perf/`, `tests/` and `docs/` linted for the first time (clean).

## Tests

- `pnpm test:ci` — utils 384, core 1814, app 167
- `pnpm format:check`, `pnpm lint`, `pnpm quality`, `docs:annotations:check`, `docs:build`
- e2e `numeric-binning` + `eat-visualization` — 43/43 against a live dev server
- docs animations `zoom`, `select-single`, `select-box`, `duplicate-badges` — 4/4, select-single now clicking real points
- `playwright test --project=screenshots` — 15/15 against a live dev server
- strict OpenSpec validation — 26 specs; four changes archived before merge per AGENTS.md
- browser-verified: the rendered row at full and 220px widths, the exact TED href, the link row's accessibility tree and focus ring, and the query builder listing labels

**Do not squash** — this touches no `apps/protspace/` path, but the repo rule stands.

https://claude.ai/code/session_01J26u3dxSKYGb8fRxchpJVa
https://claude.ai/code/session_01M9am9XhSUJdEjRLJVc7xB5


tsenoner:
### Automated review

**Does it solve #344?** Yes — the PR adds `buildTedUrl()` in `packages/core/src/components/structure-viewer/header-links.ts`, reusing the same `getBaseAccession()` + `encodeURIComponent()` normalization as the UniProt and InterPro builders, and renders a fourth `.header-link` anchor with matching `target`/`rel` attributes. It reaches real users: `apps/web/src/pages/Explore.tsx` sets `show-header="true"`, so the header renders on every protein selection, and `buildTedUrl('W6JQJ9.2')` produces exactly the URL from the issue's example. Nothing the issue asked for is missing; the only gap is peripheral documentation.

Found 1 issue:

1. Two user-facing docs pages enumerate the structure-viewer header links by name and still say AlphaFold, UniProt and InterPro, so they now contradict the shipped UI — `docs/explore/structures.md:12` and `docs/explore/index.md:60` are closed enumerations, not examples, so a reader sees a fourth link the docs say does not exist, with no explanation of what TED is (even though `docs/guide/annotations.md` already describes it). This is a one-line addition on each page; the same drift happened for the sibling feature, where the InterPro link commit (2bbff9aa) left both sentences stale and a separate docs pass (f854a36f) had to fix them four days later. The shared screenshot `docs/explore/images/structure-viewer.png` embedded on both pages also predates the change and can be regenerated with `pnpm docs:images`.

https://github.com/tsenoner/protspace/blob/ef9b3dc2ca2dc6847557f53177f8df5c4d4c188d/docs/explore/structures.md#L11-L13

Deliberately not flagged: the TED anchor rendering unconditionally (it mirrors the pre-existing AlphaFold/UniProt/InterPro anchors, and `design.md` records the decision to match existing behaviour rather than add availability checks) and the repeated anchor markup (`design.md` explicitly rejects a resource-descriptor array as scope creep here).

<sub>Reviewed at ef9b3dc2ca2dc6847557f53177f8df5c4d4c188d against issue #344.</sub>

FlorinSenoner:
Disposition: **actionable and needs to be addressed**.

Verified against current head `ef9b3dc2ca2dc6847557f53177f8df5c4d4c188d`: `structure-viewer.ts` renders TED, while `docs/explore/structures.md:12` and `docs/explore/index.md:60` still enumerate only AlphaFold, UniProt, and InterPro. Both pages embed `docs/explore/images/structure-viewer.png`; the checked-in screenshot visibly omits TED, and the documented `pnpm docs:images` pipeline regenerates it.

Required direction: update both Explore descriptions to include/link TED and regenerate the shared structure-viewer screenshot. No implementation was performed as part of this triage.

FlorinSenoner:
Implemented the verified documentation feedback in `569e67dd975c1018d43d1785a5d9ce4bcda9d708`: both Explore pages now link TED, the shared structure-viewer screenshot was regenerated and visually verified with the TED link present, and stale screenshot-pipeline readiness probes were updated for the current scatterplot/controller internals. Verification passed with `pnpm docs:images` (13 static screenshots, 7 animations, 7 GIF conversions), `openspec validate add-ted-link --strict`, and `pnpm precommit`.

tsenoner:
## Adversarial review

Reviewed in an isolated worktree by three independent lenses (code quality, adversarial correctness, issue-resolution audit), with every finding then put through a refuter whose default position was that it is a false positive. 4 raised, 2 survived refutation.

### Applied, then reverted (`7f7ca7ee`)

I initially applied the `waitForFunction` argument-position fix in `scripts/docs-screenshots/helpers.ts`,
then reverted it. An adversarial confirm pass proved it is **not** behavior-preserving, so it belongs to
you rather than to an automated cleanup pass. The branch is back to its original state; details are in
the findings below.

The triage is correct and the PR description is honest — unusually so. Nothing was narrowed,
redirected, or declined. I independently checked the two claims that could have been hand-waving:
1. "There was no TED URL builder or TED anchor, so the resource could never appear" — confirmed
against `origin/main`; `header-links.ts` had only AlphaFold/UniProt/InterPro builders.
2. The "stale image-pipeline readiness checks that blocked regeneration" claim (openspec
proposal.md:11) is real, not an excuse for scope creep. `scripts/docs-screenshots/helpers.ts:55`
used `Array.isArray(plot._plotData)`, but `_plotData` is typed `PlotData`
(`packages/core/src/components/scatter-plot/scatter-plot.ts:142`), and `PlotData` is a struct-of-
arrays *interface* with a `readonly length` field (`packages/utils/src/types.ts:125-131`) — never an
array. So `waitForDataLoad` could never return true and every screenshot capture timed out. The one-
line fix to `plot._plotData?.length` was genuinely required to regenerate `structure-viewer.png`
(which comes from `capture-static.spec.ts:207-241`, a consumer of that shared helper).
"Closes #344" is warranted. The issue is a single, fully-specified feature request with no comments
expanding its scope (I fetched it: `gh issue view 344` shows an empty comments array), the requested
URL shape is implemented exactly, and the user-visible outcome is demonstrated in the committed
screenshot. I would not downgrade this to "Related to".
The one place the PR overreached without saying so in the PR body: the `capture-animations.spec.ts`
rewrite (see gaps).

- **Scope creep in `scripts/docs-screenshots/capture-animations.spec.ts` (+47/-40) that this PR cannot verify. Only the one-line `helpers.ts:55` fix was needed to regenerate `structure-viewer.png` (produced by `capture-static.spec.ts:207`). The animations spec changes rewire the duplicate-stack probes from the removed `_duplicateStackByKey` / `_expandedDuplicateStackKey` to `_dupOverlay.byKey` / `_dupOverlay.expandedKey`, add a forced PCA-projection selection that throws if absent, and switch a `waitForFunction` to explicit `polling: 200`.**
  - _Why it matters:_ These probes drive the duplicate-badge GIF, which this PR does not regenerate — the only image changed is `structure-viewer.png`. The new field paths are runtime-reachable (TypeScript `private` is soft-private; I confirmed `byKey` at `duplicate-stack-overlay-controller.ts:79` and `expandedKey` at :81, and `_dupOverlay` at `scatter-plot.ts:241`), so the fix is plausible — but it ships unexercised. If the projection-index or polling change is subtly wrong, the breakage surfaces on some future unrelated docs regeneration, and the blame will point at that PR rather than this one. It also makes this PR's diff larger than its stated purpose.
  - _Suggested follow-up:_ Either split the `capture-animations.spec.ts` repair into its own PR that actually runs the animation captures and commits the regenerated GIF, or run the duplicate-badge capture once on this branch and note the result in the PR body so the change is not merged untested.
- **TED coverage is narrower than UniProt/InterPro, but the link is rendered unconditionally for every `proteinId`. `getBaseAccession` only strips a `.`-suffix, so an isoform ID like `P12345-2` or a non-UniProt dataset ID (`my_seq_001`, `sp|P12345|NAME`) is passed straight through and produces a dead TED URL.**
  - _Why it matters:_ `docs/guide/faq.md:50` already tells users "Structures load automatically from AlphaFold if your protein IDs are UniProt accessions", i.e. non-UniProt datasets are an expected case. TED is derived from the AlphaFold DB, so its per-accession coverage is strictly narrower than UniProt's and somewhat narrower than InterPro's — the dead-link rate for the new TED link will be visibly higher than for the two links it sits beside. The design doc (design.md:37) consciously accepts this ("let the external service report availability"), which is defensible and consistent with existing behavior, so this is a follow-up rather than a blocker.
  - _Suggested follow-up:_ Open a follow-up to handle unresolvable resource links uniformly — e.g. suppress or visually de-emphasize the UniProt/InterPro/TED row when `_loadStructure` reports `No 3D structure was found for <id>` (structure-viewer.ts:182), since that error is already a strong signal the accession is not in the AlphaFold DB and therefore not in TED.
- **`.header-links` is `display: flex` with **no** `flex-wrap` (`packages/core/src/components/structure-viewer/structure-viewer.styles.ts:73-77`); only the outer `.header-info` sets `flex-wrap: wrap` (:61-66). The design doc (design.md:39) states the change "Reuse[s] the existing wrapping flex container", which is inaccurate about the container the three links actually live in.**
  - _Why it matters:_ The link row grew from two labels to three plus an extra separator and can now only wrap as a whole unit, not between links. At the app's default right-panel width it fits — the regenerated screenshot proves that — but a narrower panel or a longer accession would overflow the row horizontally rather than wrap. There is no regression test or screenshot covering a narrow/mobile layout, so this would only be caught by eye.
  - _Suggested follow-up:_ Add `flex-wrap: wrap` to the `.header-links` rule (a one-line, zero-risk change), and correct the design.md sentence so the accepted trade-off is recorded against the container that actually applies.
- **`docs/explore/control-bar.md:39` still describes the annotation dropdown sections as "(UniProt, InterPro, Taxonomy, Other)", while `packages/core/src/components/control-bar/annotation-categories.ts:9` defines `'Biocentral' | 'InterPro' | 'TED' | 'Taxonomy' | 'UniProt' | 'Other'`.**
  - _Why it matters:_ Pre-existing staleness, not introduced here and genuinely outside issue #344's scope (it concerns annotation grouping, not resource links). Flagging it only because it is a TED-specific doc gap that this PR walked directly past while updating the two other TED-mentioning Explore docs — a reader comparing the docs will see TED acknowledged in one place and missing in another.
  - _Suggested follow-up:_ Fold the one-line list update into this PR (it is already touching `docs/explore/`) or file a small `docs:` issue to resync `control-bar.md` with `annotation-categories.ts`.
- **The OpenSpec change is not archived: the new `protein-resource-links` capability exists only under `openspec/changes/add-ted-link/specs/`, never merged into `openspec/specs/` (which currently holds only `bundle-format-contract`, `prep-failure-routing`, `prep-observability`, `support-contact`).**
  - _Why it matters:_ AGENTS.md states `openspec/specs/` is "the source of truth for current behavior", so until `/opsx:archive` runs, the shipped resource-link behavior is not represented in the source of truth. Low severity: five other changes in the repo (`add-eat-visualization`, `unified-visibility-model`, `optimize-e2e-suite`, …) are likewise unarchived, so the repo's actual practice is clearly to archive on a separate cadence — this is consistent, not a deviation.
  - _Suggested follow-up:_ No action needed for this PR; run `/opsx:archive add-ted-link` in the next batch archive pass so `protein-resource-links` lands in `openspec/specs/`.

</details>

#### 1. `waitForDataLoad`/`waitForLegend` pass their `waitForFunction` options in the `arg` position, so the `timeout` and `polling` they compute are silently discarded.

`scripts/docs-screenshots/helpers.ts:55` · **low** · correctness · _applied then reverted_

Playwright's signature is `waitForFunction(pageFunction, arg, options)` (verified in `playwright-
core@1.61.0/types/types.d.ts:5880`) — there is no arity sniffing. `helpers.ts:50-60`, `:64-67` and
`:78-86` pass `{ timeout, polling: 200 }` as the second positional argument, so it is serialized as
the (unused) page-function argument and the real options object is `undefined`. This PR fixed
exactly this mistake in the sibling file (`capture-animations.spec.ts:564-575` now passes
`undefined, { timeout: 10_000, polling: 200 }`) but left it in the call it edited. Concrete failure:
`await waitForDataLoad(page, 5000)` does not fail after 5s — `actionTimeout` is unset in
`playwright.config.ts`, so the wait runs unbounded until the 60s test timeout and reports an opaque
"Test timeout of 60000ms exceeded" instead of a `waitForFunction` failure naming the readiness
predicate. This is exactly how the now-fixed `Array.isArray(plot._plotData)` staleness manifested.
`polling: 200` is likewise ignored (defaults to `raf`).

**Why it was reverted rather than kept:** `playwright.config.ts` leaves `actionTimeout` unset (default
`0`), so all three waits were previously *unbounded*, capped only by the 60s per-test timeout. Applying
the fix caps them at 30s/30s/15s and moves polling from `raf` to fixed 200/100/200ms intervals. A probe
spec confirmed the old form never times out while the new form rejects on schedule. That is almost
certainly what you intended when you wrote `{ timeout, polling }` — but it can turn a slow-but-successful
screenshot capture into a hard failure, so it is your call, not an automated one. Blast radius is small:
`docs:screenshots`/`docs:animations`/`docs:images` are referenced by no workflow in `.github/workflows/`,
so this runs only on manual local invocations.

<details><summary>The fix, if you want it</summary>

In /Users/tsenoner/Documents/projects/protspace-suite/.protspace-wt-pr402/scripts/docs-
screenshots/helpers.ts, move the options object into the third argument position by inserting
`undefined,` before it at the three call sites named in the finding:
1. line 50-60 (waitForDataLoad readiness predicate): change the trailing `},\n    { timeout,
polling: 200 },\n  );` to `},\n    undefined,\n    { timeout, polling: 200 },\n  );`
2. line 64-67: `await page.waitForFunction(() => !document.getElementById('progressive-loading'),
{\n    timeout,\n    polling: 100,\n  });` becomes `await page.waitForFunction(\n    () =>
!document.getElementById('progressive-loading'),\n    undefined,\n    { timeout, polling: 100 },\n
);`
3. line 78-86 (waitForLegend predicate): change `},\n    { timeout, polling: 200 },\n  );` to `},\n
undefined,\n    { timeout, polling: 200 },\n  );`
This is exactly the form already used at capture-animations.spec.ts:564-575 and at helpers.ts:100
(waitForWebGLContext), so it type-checks. Note for a follow-up (do not autofix in this pass):
helpers.ts:141, :156 and :175 have the identical misplacement, and fixing :156/:175 is what restores
their intended `try/catch` fallbacks; the same pattern also appears in capture-static.spec.ts.

</details>

#### 2. Forcing `plot.selectedProjectionIndex` directly bypasses the control bar, so the recorded duplicate-badges.gif shows a projection dropdown that contradicts the plotted data.

`scripts/docs-screenshots/capture-animations.spec.ts:554` · **medium** · correctness

Concrete failure: `apps/web/public/data.parquetbundle` (the dataset `/explore` loads) has
projections `['ProtT5 — UMAP 2', 'ProtT5 — PCA 2', 'ESM2-650M — UMAP 2', 'ESM2-650M — PCA 2']`, and
`data-renderer.ts:94` starts the plot at index 0. The new code computes `pcaIndex = 1` and assigns
`plot.selectedProjectionIndex = 1` on the scatterplot element. The control bar only mirrors that
value inside `_syncWithScatterplot()`, which `control-bar.ts:1233-1237` schedules exactly once via
`setTimeout(..., 50)` at auto-sync setup; a `selectedProjectionIndex` change does not dispatch
`data-change` (`scatter-plot.ts:764-782` only emits on `data` / `filteredProteinIds` /
`filtersActive`), so nothing re-syncs it. Result: the plot renders `ProtT5 — PCA 2` while the
control-bar trigger (`control-bar.ts:564`, which renders `this.selectedProjection`) still reads
`ProtT5 — UMAP 2`. The animation project records the full 1536x864 viewport
(`playwright.config.ts`), so the control bar is in frame and the published
`docs/explore/images/duplicate-badges.gif` would ship that contradiction. The switch also appears
unnecessary: `duplicate-badges.gif` was successfully generated on the default projection in 68c3bd91
(2026-05-11), before the `_duplicateStackByKey` rename broke the probe, and duplicate grouping is
cross-projection (0c6e0794). The GIF is not regenerated in this PR, so this path is unverified.

In /Users/tsenoner/Documents/projects/protspace-suite/.protspace-wt-pr402/scripts/docs-
screenshots/capture-animations.spec.ts, replace the body of the first `page.evaluate` in the
`duplicate-badges.gif` test (lines 542-558) with a control-bar-driven selection so the label and the
plotted data stay in sync:
```ts
await page.evaluate(async () => {
  const plot = document.querySelector('#myPlot') as DuplicatePlotProbe | null;
  const controlBar = document.querySelector('#myControlBar') as
    | (HTMLElement & { applyProjectionSelection(projection: string): void })
    | null;
  if (!plot || !controlBar) {
    throw new Error('Duplicate-badge capture needs #myPlot and #myControlBar');
  }
  const pca = plot.data?.projections?.find((projection) => projection.name.includes('PCA'));
  if (!pca) {
    throw new Error('Duplicate-badge capture requires a PCA projection');
  }
  controlBar.applyProjectionSelection(pca.name);
  await plot.updateComplete;
  plot.config = { ...(plot.config ?? {}), enableDuplicateStackUI: true };
  await plot.updateComplete;
});
```
`applyProjectionSelection` (control-bar.ts:139) sets `selectedProjection`, mirrors the index onto
the scatterplot while `autoSync` is on (data-renderer.ts:209 turns it on after load), and dispatches
`projection-change` so the app's view controller/URL stay consistent.
`DuplicatePlotProbe.selectedProjectionIndex` can then be dropped from the type at line 17 if unused.
Do not autofix: the capture must be re-run and duplicate-badges.gif regenerated to confirm PCA still
yields a multi-point stack before this is trusted.

</details>

_2 further finding(s) were raised and refuted during verification._

FlorinSenoner:
Addressed the adversarial review in [comment 5203113387](https://github.com/tsenoner/protspace/pull/402#issuecomment-5203113387) on `52b8afc3f40313e6dd6377a4a510af5ef292b9ab`.

1. **Animation-script scope:** kept. `pnpm docs:images` is the documented canonical image pipeline and includes the animation project. A live probe of the current bundled dataset found 0 multi-point stacks on the default ProtT5 UMAP projection and 265 on ProtT5 PCA, so the PCA switch is required for this capture rather than incidental scope. The focused duplicate-badge capture now passes and the GIF was regenerated.
2. **TED availability for non-UniProt IDs:** no code change. `openspec/changes/add-ted-link/design.md` explicitly makes availability checks a non-goal and records deterministic-link parity with UniProt/InterPro; changing that policy would be a separate product change.
3. **Header wrapping:** no code change. `.header-info` is the existing wrapping container named by the design, while `.header-links` contains only fixed link labels; a longer accession is a separate `.protein-id` flex item and does not widen the link row. Styling changes are also an explicit non-goal for this change.
4. **Control-bar category docs:** no change. The stale annotation-category list predates this PR and concerns annotation grouping, not the resource-link capability in #344.
5. **OpenSpec archival:** no change. The change remains active while the PR is open; archive/merge into `openspec/specs/` belongs to the post-merge archive pass.
6. **`waitForFunction` options:** fixed at the three reviewed call sites by passing `undefined` as the page-function argument. Added real Playwright regressions for readiness, loading-overlay, and legend timeouts. RED: each 100 ms timeout ran until the 2,000 ms test timeout; GREEN: all 3 pass.
7. **Projection/control-bar mismatch:** fixed by selecting PCA through `controlBar.applyProjectionSelection()` and asserting the plotted projection equals the visible control-bar selection before recording. RED: the assertion timed out with direct plot mutation; GREEN: the complete 21-action duplicate-badge capture passed, and the regenerated GIF visibly labels `ProtT5 — PCA 2`.

Verification: focused helper tests `3/3`; focused duplicate-badge Playwright capture `1/1`; GIF conversion `1/1`; `openspec validate add-ted-link --strict`; staged `pnpm precommit` (ESLint/Prettier, typecheck, Knip, annotation-doc check, docs build). Exact-head checks: Code Quality passed, Build Documentation passed, Bundle format contract passed, E2E path-filter skipped.

Issue #424, opened 2026-08-10
[FEATURE] EAT audit mode: flag likely-miscurated annotations from neighbour disagreement


tsenoner:
### Feature Request

Today `protspace transfer` only runs in one direction: it fills annotations that are **missing**. Add the reverse direction — run EAT over proteins that **already carry a label** and surface the ones whose nearest neighbours in embedding space confidently disagree, as candidate curation errors.

This is the "errors out" half of the transfer story: the same kNN operation that fills a gap also audits the known.

### Why this matters now

The manuscript's headline audit result is currently computed in the **wrong space**. `sections/03.results.md` derives ~860 candidate misannotations from *"fewer than half of their 50 nearest **UMAP** neighbors share the same annotation"*, while `sections/05.methods.md` states that finding neighbours in the 2-D projection rather than the full embedding space is *"a correctness requirement rather than a performance choice"* — non-linear DR is not an isometry. The cover-letter notes already self-flag this as a liability. This feature is what makes that result defensible.

### Proposed shape

`protspace transfer --mode {fill,audit,both}` (default `fill`, byte-identical to today) rather than a new command — bundle I/O, id-column disambiguation, column validation and format-version migration are already there and would otherwise be copy-pasted. `--mode both` runs a single kNN pass and emits both column families, which matters because the GEMM dominates the cost.

**Leave-one-out is the crux.** `protlabel.eat()` gains `exclude_ids`, aligned positionally with `query_ids`; every reference row whose id equals the query's id is dropped before reliability accumulation. It must key on **identifier, not distance**:

- Not distance — a `d ≈ 0` rule deletes exactly the legitimate duplicate-embedding neighbours, which are the strongest evidence, and cosine self-distance is not bit-exactly 0 (`clip(1 - cos, 0, 2)` computed float64-from-float32).
- Not a single row index — `ref_ids` is not unique; `run_transfer` builds it without a uniqueness check, so one protein can occupy several reference rows.

`eff_k` therefore becomes per-query (`min(k, n_refs - self_count_i)`); a query with zero survivors emits no prediction rather than an error. With `exclude_ids=None` results stay bit-identical to today.

### Scoring statistic

Report the **reliability-index margin**, not a raw confidence:

```
disagreement = clamp(RI(best_disagreeing_label) - RI(curated_label), 0, 1)
support      = RI(curated_label)          # emitted alongside, never folded in
```

[...]

PR #425, opened 2026-08-10
feat(eat): self-transfer, reliability filter modes, and glyph outline parity

### LINKED ISSUE #369: EAT glyphs: reconcile edge-darkening border between predicted rings and filled dots
## Context

In the WebGL point fragment shader (`packages/core/src/components/scatter-plot/webgl/renderer/export-shaders.ts`, `POINT_FRAGMENT_SHADER`), regular **filled** points get a subtle edge-darkening border/outline:

```glsl
// Darken near the edge to mimic a border/outline.
float strokeWidth = 0.15;
if (v_predicted < 0.5 && v_color.a > 0.5 && max(edgeDist, 0.0) < strokeWidth) { …darken near edge… }
```

Predicted **EAT** glyphs (`v_predicted > 0.5`) intentionally **skip** this border — they render as hollow rings instead (added in PR #315). So the two point types are visually distinguished, but the border treatment is inconsistent: filled dots have a darker outline; EAT rings do not.

## Options

1. **Add the same edge-darkening to EAT glyphs** for a consistent outline. Caveat: EAT glyphs are hollow rings, so an outer edge-darken tends to thicken/muddy the ring edge; they're already distinguished by being hollow + a thicker stroke.
2. **Remove the border from all dots** for the opposite consistency. This is a broader, app-wide visual change (every dataset, not just EAT) — the border currently gives filled dots definition/depth.
3. **Keep as-is** — filled dots keep the subtle border; EAT rings stay borderless (hollow is their distinguisher).

## Rendering-time note (do not chase this as a perf win)

The border is a per-fragment conditional darken (a branch + a `mix` near the sprite edge) — a handful of ALU ops per fragment. On the GPU this is **not** a rendering bottleneck; the scatter plot is dominated by vertex throughput and overdraw at 570k+ points, not fragment ALU. **Removing the border would not measurably improve rendering time.** Decide this on visual grounds, not performance.

---
Surfaced during PR #315 (EAT visualization) review. Not blocking that PR.

### LINKED ISSUE #380: [BUG] Make the EAT reliability filtering work for all kind of different settings
### Bug Description

smaller then, latrger then, and between and sync between filter and drag

### Code of Conduct

### LINKED ISSUE #393: [FEATURE] Apply EAT to all NA values in the current dataset
### Feature Request

Instead of needing to provide a separate dataset to apply EAT on, add a feature to apply it to annotations containing NA values.

### Code of Conduct

tsenoner:
Closes #393. Closes #380. Closes #369. Groundwork for #424.

Four EAT items needed for the Nature Methods submission. Each turned out to be materially larger than its issue text, so the findings are written up here: **[implementation plan](https://claude.ai/code/artifact/e4a7b78b-b95d-4cf2-9a2b-d7cbe81cb501)**.

> **Do not squash-merge** — this touches `apps/protspace/`. Squashing collapses the branch into one commit touching every path and destroys semantic-release's per-commit path scoping.

---

## #393 — EAT within one dataset

An empty `Rule` matched nothing rather than everything, so running `protspace transfer` without `--query-*`/`--reference-*` raised *"Classifier matched no query proteins"*. Users had to invent a column describing "the rows that are empty" — producing the shipped EAT demo bundle needed exactly that, a synthetic `eat_split` column plus two flags.

An empty rule now reads as **open**. Self-transfer needed no new transfer logic: `run_transfer` already split queries from references *per column* on missing-vs-present, and those sets are exact complements of `_is_missing`, so **a protein can never be its own reference**.

Making empty rules match everything was **not sufficient on its own**, which is the part worth reviewing. `classify()` partitioned with `if`/`elif`, so two open rules handed every protein to the query list and left references empty — transfer would then skip every column and write the bundle back unchanged **at exit 0**. Precedence now applies only between two *explicit* rules.

That also fixes two pre-existing silent no-ops: passing only `--query-*`, and the `--query-where 'col~'` match-everything workaround.

**Verified on the 832-protein phosphatase fixture**, with no rules at all:

| Column | Held-out | Exact | Manuscript (`05.methods.md:95`) | Self-sourced |
|---|---|---|---|---|
| `ec` | 213 | 91.55% | 91.55% | 0 |
| `protein_families` | 213 | 99.06% | 99.06% | 0 |

## #380 — the reliability filter

The legend slider identified its condition by pattern-matching `NOT(conf < X)` on operator *and* logical op, at the top level only. That is the shape the UI is least likely to produce: picking an eat-confidence column in the query builder creates a condition defaulting to `gt`. So the most natural gesture left the slider reading 0% while the plot was filtered, and the next drag **appended** a contradictory condition instead of replacing it.

- Conditions are now identified by **ownership of the eat-confidence column**, at any depth, tagged or not — so a hand-built condition mirrors instead of being duplicated, and one restored from an older bundle still works.
- The control gains **three modes** (`Hide below` / `Hide above` / `Keep between`), each expressed as NOT-form conditions. Only the NOT index-complement re-includes null-confidence rows, so **curated proteins stay visible in every mode by construction** — the slider's help text promised exactly that, and it was false for three of the five reachable shapes. The evaluator's deliberate null semantics are untouched.
- Deriving the state back respects the logical op: `NOT(conf < X)` is a lower bound, a bare `conf < X` is an upper one. Reading a bare `lt` as a lower bound made the slider skip a replacement and leave a contradiction in place — caught by a regression test, not by inspection.
- The mirror's de-dupe guard was one shared scalar, so a drag on one base was compared against another's; it is now keyed per column, and switching the coloured-by annotation force-emits.
- A query matching nothing was pushed as an **active** filter, which reads as "hide everything" and blanked the canvas with no way back. It now clears instead.

## #369 — glyph outlines

Neither option in the issue is right. Adding the existing darkening to rings would consume 27–50% of the annulus at every size, smearing the hollow cue; removing it everywhere costs dense clusters their only separator between adjacent same-hue dots.

The real defect is the mechanism. The band was a **fraction of the sprite** (`0.15`), so its absolute width shrank with point size and export scale — 0.35px at `gl_PointSize` 48, 1.13px at 512. At the app's own `nature-1col` preset (89mm @ 300dpi) that put the observed-dot outline at ~0.24pt of tint, **below the reproducible-line floor**, while the predicted ring at 0.49pt solid survived. In a published figure only one glyph class had an outline at all.

The outline is now measured in **device pixels**, derived from the isotropic `pixelScale` the ring already used, applied to both classes and budget-capped on the ring. Filled-vs-hollow remains what tells them apart. The inner edge is anti-aliased; it was a hard threshold on a field whose outer edge was already smoothed.

## #424 — audit mode (filed, not implemented)

The manuscript's headline audit result is computed in the space its own Methods section forbids: `03.results.md:37` derives ~860 candidate misannotations from *"50 nearest **UMAP** neighbors"*, while `05.methods.md:73` states that using the 2-D projection rather than the embedding space is *"a correctness requirement rather than a performance choice"*. `COVER_LETTER_NOTES.md` already self-flags this, and `STORY_v2` stages that panel as **Figure 2C, the paper's climax**.

#424 captures the design — `--mode audit`, leave-one-out keyed on identifier rather than distance, a reliability-margin score, a distinct `__audit_*` column family (reusing `__pred_*` would let `materializeEatOverlay` overwrite curated biology) — plus the prior art that must be cited, including **Heinzinger 2022, already cited in the manuscript, which does this and calls it essential for cleaning up databases**, and **CLEAN (Science 2023)**, which is not from this lab.

---

## Verification

- `uv run pytest` — **840 passed**; `uv run ruff check` clean
- `pnpm test:ci` — core **1551**, utils **371**, app **165**; `pnpm format:check` clean; `pnpm precommit` on every commit
- Playwright `eat-visualization` against the real phosphatase bundle — proves the shader **compiles** (string tests cannot) and the legend layout holds at 320/390/601/800px
- `openspec validate --strict` passes

Every new test was watched failing first. For #393 the tests were re-run against the pre-change `classify()` to confirm all five fail there and pass here.

## Deliberately not in this PR

Audit mode (#424); persisting the reliability *mode* to bundle settings (schema addition needing a migration story for the legacy scalar); export legend renderers drawing EAT rings, so exported figures explain the encoding; and an EAT docs page — `transfer` is only name-checked at `python-cli.md:125`, and NM's code checklist wants install time, demo run time and expected output.

Three pre-existing defects the shader audit surfaced, unrelated to the border question and worth their own issues: `edgeDist` is not gradient-normalised so glyphs are half as thick on diamonds; `u_knockoutColor` is dead but still costs a forced `getComputedStyle()` per frame; and the Observed/Predicted key never reaches any exported figure.

https://claude.ai/code/session_016tSUUWsGomA4BFoVFGvSwj

### DIFF
--- a/apps/protspace/src/protspace/analysis/classification.py
+++ b/apps/protspace/src/protspace/analysis/classification.py
@@ -1,8 +1,14 @@
 """Classify proteins as transfer queries vs annotated references.
 
 Rules match by ID prefix and/or a case-insensitive metadata substring
-(``column ~ substring``). No biology is hardcoded; a query rule that matches
-nothing is an error.
+(``column ~ substring``). No biology is hardcoded; an *explicit* query rule
+that matches nothing is an error.
+
+A rule with no clauses is **open** — it restricts nothing and matches every
+protein. Passing no rules at all therefore means "transfer within this
+dataset": every protein is a candidate on both sides, and the caller splits
+them per column into the ones missing a value (queries) and the ones holding
+one (references).
 """
 
 from __future__ import annotations
@@ -14,13 +20,30 @@ import pyarrow as pa
 
 @dataclass
 class Rule:
-    """A classification rule. A protein matches if ANY clause matches."""
+    """A classification rule. A protein matches if ANY clause matches.
+
+    A rule with no clauses at all is *open*: it expresses "no restriction"
+    rather than "match nothing" (see ``is_open``).
+    """
 
     id_prefixes: list[str] = field(default_factory=list)
     where: list[tuple[str, str]] = field(default_factory=list)  # (column, substring)
 
+    @property
+    def is_open(self) -> bool:
+        """True when the rule carries no clauses, i.e. it restricts nothing.
+
+        An open rule matches every protein. This is what makes self-transfer
+        (#393) work without a marker column: with both rules open, every
+        protein is a candidate query *and* a candidate reference, and
+        ``run_transfer`` does the real split per column on missing-vs-present.
+        """
+        return not self.id_prefixes and not self.where
+
 
 def _matches(rule: Rule, identifier: str, column_data: dict[str, list], i: int) -> bool:
+    if rule.is_open:
+        return True
     if any(identifier.startswith(p) for p in rule.id_prefixes):
         return True
     for column, substring in rule.where:
@@ -39,8 +62,15 @@ def classify(
 ) -> tuple[list[int], list[int]]:
     """Return (query_indices, reference_indices) into the annotations table.
 
-    Query classification takes precedence: a protein matching both rules is a
-    query. Raises ValueError if the query rule matches nothing.
+    Between two *explicit* rules, query classification takes precedence: a
+    protein matching both is a query and never a reference. An **open** rule
+    (no clauses) expresses no preference, so it never takes a protein away from
+    the other set — which is what lets both sets cover the whole table when no
+    rules are given. The two lists may therefore overlap; callers are expected
+    to make the final, disjoint split themselves (``run_transfer`` does it per
+    column on missing-vs-present).
+
+    Raises ValueError if an explicit query rule matches nothing.
 
     ``identifiers`` may be a pre-materialized list of the string identifier
     column (same order as ``annotations``) to avoid re-materializing it when the
@@ -63,15 +93,30 @@ def classify(
     }
     column_data = {c: annotations.column(c).to_pylist() for c in where_columns}
 
+    # Precedence applies only *between two explicit rules*: there, a protein
+    # matching both is a query and never a reference. When either side is open
+    # it expresses no preference, so it must not consume proteins away from the
+    # other set — otherwise two open rules would hand every protein to the
+    # query list and leave references empty, and transfer would silently
+    # do nothing.
+    both_explicit = not query_rule.is_open and not reference_rule.is_open
+
     query_indices: list[int] = []
     reference_indices: list[int] = []
     for i, identifier in enumerate(identifiers):
-        if _matches(query_rule, identifier, column_data, i):
+        is_query = _matches(query_rule, identifier, column_data, i)
+        is_reference = _matches(reference_rule, identifier, column_data, i)
+        if is_query and is_reference and both_explicit:
+            is_reference = False
+        if is_query:
             query_indices.append(i)
-        elif _matches(reference_rule, identifier, column_data, i):
+        if is_reference:
             reference_indices.append(i)
 
     if not query_indices:
+        if query_rule.is_open:
+            # Nothing to blame on the rules — the table itself is empty.
+            raise ValueError("Classifier found no proteins to use as queries.")
         raise ValueError(
             "Classifier matched no query proteins; check --query-id-prefix / "
             "--query-where rules."
--- a/apps/protspace/src/protspace/cli/transfer.py
+++ b/apps/protspace/src/protspace/cli/transfer.py
@@ -131,8 +131,10 @@ def run_transfer(
 
     if total_transferred == 0:
         logger.warning(
-            "No annotations were transferred. Check the --reference-* rules and "
-            "that query proteins have missing values in the target column(s)."
+            "No annotations were transferred: the target column(s) have either "
+            "no missing values to fill or no values to transfer from. If you "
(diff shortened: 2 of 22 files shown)


tsenoner:
## ⚠️ Conflicts with #416 — converting to draft

Flagged before this goes further: the #380 work in this PR **conflicts fundamentally with #416**, not just textually.

### The conflict

This PR's reliability filter is built on the premise that *"every mode is a `NOT`, so curated points are retained by the index complement."* #416 removes that premise:

> **`NOT` is no longer a bare set complement.** It now means *"has a value **and** does not match"*

Under #416, every `NOT(confidence …)` condition this PR emits would **hide every curated protein** — the exact outcome both PRs are trying to prevent. #416 already saw this and solved it the other way:

> It is now a single un-negated condition — `confidence >= X` carrying an N/A chip

#416 also adds the `gte`/`lte` operators this PR deliberately did without, and rewrites `_isReliabilityConditionForKey` to the new shape.

**Files touched by both:** `control-bar.ts`, `query-types.ts`, `control-bar.query-apply.test.ts`.

### What is unaffected

- **#393** (`apps/protspace/` — `classification.py`, `transfer.py`) — #416 touches no Python.
- **#369** (`export-shaders.ts`) — #416 touches no shader.

Verified: `git diff --name-only main...feat/improved_filtering` contains no Python and no WebGL files.

### What survives a rebase onto #416

These are genuinely additive — #416 does none of them:

- Ownership-based condition matching **at any depth** (#416 still shape-matches at the top level only, so a condition nested in a group is still invisible to it)
- Per-base mirror de-dupe guard, replacing the shared scalar
- Force-emit on annotation switch
- Empty match set **clearing** the filter instead of blanking the canvas
- The legend's three-mode control (`Hide below` / `Hide above` / `Keep between`)

The mode control actually lands *more cleanly* on #416's model — each mode becomes one un-negated condition with an N/A presence chip (`>= X`, `<= X`, `between A,B`) instead of one or two negated ones.

### Proposed

Split: land #393 + #369 (independently verified, zero overlap), and rebase the #380 work onto #416 — stacked the way #417 is — rewriting `eat-reliability.ts` to emit #416's shapes.

tsenoner:
Superseded by #426 (#393 + #369, no overlap with #416) — see the conflict analysis above. The #380 reliability-filter work is being restacked on `feat/improved_filtering` so it builds on #416's N/A-aware `NOT` instead of the bare complement it currently assumes.

PR #426, opened 2026-08-10
feat(eat): transfer within one dataset, and one outline treatment for both glyph classes

### LINKED ISSUE #369: EAT glyphs: reconcile edge-darkening border between predicted rings and filled dots

### LINKED ISSUE #393: [FEATURE] Apply EAT to all NA values in the current dataset

tsenoner:
Closes #393. Closes #369.

Two EAT items needed for the manuscript submission. Both are independent of the filter-query work in #416/#417 — this branch touches no query-builder file.

Full write-up of the research behind all four EAT items: **[implementation plan](https://claude.ai/code/artifact/e4a7b78b-b95d-4cf2-9a2b-d7cbe81cb501)**.

> **Do not squash-merge** — this touches `apps/protspace/`. Squashing collapses the branch into one commit touching every path and destroys semantic-release's per-commit path scoping.

> Supersedes #425, which also carried the #380 reliability-filter work. That part conflicts semantically with #416 (which redefines `NOT` from a bare set complement to "has a value **and** does not match") and is being restacked on `feat/improved_filtering` instead.

---

## #393 — EAT within one dataset

An empty `Rule` matched nothing rather than everything, so running `protspace transfer` without `--query-*`/`--reference-*` raised *"Classifier matched no query proteins"*. Users had to invent a column describing "the rows that are empty" — producing the shipped EAT demo bundle needed exactly that, a synthetic `eat_split` column plus two flags.

An empty rule now reads as **open**. Self-transfer needed no new transfer logic: `run_transfer` already split queries from references *per column* on missing-vs-present, and those sets are exact complements of `_is_missing`, so **a protein can never be its own reference**.

**The part worth reviewing:** making empty rules match everything was *not sufficient on its own*. `classify()` partitioned with `if`/`elif`, so two open rules handed every protein to the query list and left references empty — transfer would then skip every column and write the bundle back unchanged **at exit 0**. Precedence now applies only between two *explicit* rules.

That also fixes two pre-existing silent no-ops: passing only `--query-*`, and the `--query-where 'col~'` match-everything workaround.

**Verified on the in-repo 832-protein phosphatase fixture**, with no rules at all:

| Column | Held-out | Exact | Manuscript (`05.methods.md:95`) | Self-sourced |
|---|---|---|---|---|
| `ec` | 213 | 91.55% | 91.55% | 0 |
| `protein_families` | 213 | 99.06% | 99.06% | 0 |

The command drops from this, plus a hand-built `eat_split` column:

```
protspace transfer -b … -e … -t ec -t protein_families -o … \
  --query-where eat_split~query --reference-where eat_split~reference
```

to just `protspace transfer -b … -e … -t ec -t protein_families -o …`.

### The mirror-image no-op, also fixed

Review of the above turned up the same defect on the half nobody had fixed. An explicit *query* rule matching nothing always errored; an explicit *reference* rule matching nothing returned an empty set, which `run_transfer` turned into "no references with a value; skipping" for every column — so a typo'd `--reference-id-prefix` wrote the bundle back unchanged and **exited 0**.

Both sides now error. This is not a new CLI contract: `run_transfer` already routes `ValueError → typer.BadParameter`, so it surfaces exactly like the query-side error (exit 2). It also covers the case where two explicit rules overlap so completely that precedence empties the reference set — the message names the query rule as the likely cause.

## #369 — glyph outlines

Neither option in the issue is right. Adding the existing darkening to rings would consume 27–50% of the annulus at every size, smearing the hollow cue; removing it everywhere costs dense clusters their only separator between adjacent same-hue dots.

The real defect is the mechanism. The band was a **fraction of the sprite** (`0.15`), so its absolute width shrank with point size and export scale — 0.35px at `gl_PointSize` 48, 1.13px at 512. At the app's own `nature-1col` preset (89 mm @ 300 dpi) that put the observed-dot outline at ~0.24 pt of tint, **below the reproducible-line floor**, while the predicted ring at 0.49 pt solid survived. In a published figure only one glyph class had an outline at all.

The outline is now `max(fraction, device-pixel floor)` — the fraction keeps the rim growing with the glyph on screen, the floor keeps it from going sub-pixel in print — applied to both classes and capped by a per-class budget. Filled-vs-hollow remains what tells them apart: a pre-attentive categorical difference, far stronger than outline weight. The inner edge is anti-aliased; it was a hard threshold on a field whose outer edge was already smoothed.

### Two defects found in review of that change

- **The device-pixel floor was unbounded on filled dots.** It is `2/gl_PointSize` in field units, so it grows without limit as sprites shrink: at the app's default point size roughly two-thirds of every observed dot was mixed toward `finalColor * 0.5`, so dense or zoomed-out views — the 570K-protein case ProtSpace is built for — rendered points markedly darker than the hue their legend swatch shows. `OUTLINE_DOT_BUDGET` now caps it, alongside the ring's existing budget. Both are resolved into one `outlineBudget` inside the branch that already interprets the glyph class, so the outline code downstream is class-free.
- **The diamond got half a pixel.** Converting device pixels into `edgeDist` units assumed `|grad edgeDist| == 1`, but the diamond's field has `|grad| = 2`. Its "one device pixel" therefore rendered at 0.5 px — the same sub-visible-line defect #369 exists to fix, relocated from one glyph class to one glyph shape. The conversion now goes through the field's own L2 gradient (not `fwidth`, whose L1 norm would reintroduce the diagonal anisotropy the ring depends on avoiding), clamped to the true gradient range across all six shapes.
- **…and fixing that width briefly broke the feather.** The smoothstep span was widened along with the width. On a ring `outlineBudget` caps the band *below* one device pixel (`ringWidth * 0.35 ≤ 0.19`), so a `fieldPerPixel`-wide ramp is wider than the band it softens and washes the rim out instead. The diamond took the full 2×: at the app's default `pointSize` 240 a predicted diamond's darkening at the glyph edge fell from 0.755 to 0.266. The feather is back on `pixelScale` while the width keeps `fieldPerPixel` — widening the band is safe, widening the feather past the band is not. Every other shape has gradient 1, so the two are equal there and the expression is byte-identical to what shipped.

That last one was invisible in the reference screenshot because the harness pushed shape `0` for every glyph — it drew circles only, and the diamond is precisely where the maths differs. It now renders shape × size rows, skewed small where the device-pixel floor dominates.

## Docs

Only the inline `--help` text had been updated, so every doc a user actually reads still taught the workaround:

- `docs/cli.md` — scope paragraph, no-filter example first, and the Default column filled in (it listed `—` for all four filters).
- `ProtSpace_Transfer.ipynb` — the default cell no longer passes filters; the narrowing example moved to its own markdown cell.
- `ProtSpace_Preparation.ipynb` — dropped the two pre-flight refusals that returned early unless both filters were set, and relabelled the widgets, which still said "(at least one)".

## Test infrastructure

The `glyph-outline` visual harness said "not a CI gate" in its own header but was added to `playwright.config.ts` unconditionally, and `pnpm test:e2e` runs the config with no `--project` filter — so it was in the default selection on every PR touching `apps/web` or `packages/`. It asserts nothing beyond the shader compiling, and its only product is a gitignored PNG that is not among the workflow's artifact upload paths, so CI rendered it and discarded it. It is now gated behind `RUN_GLYPH_HARNESS=1`, and all three env-gated projects share one `optIn()` helper.

---

## Verification

- `uv run pytest` from `apps/protspace/` — **806 passed**, 2 skipped; `uv run ruff check` + `ruff format --check` clean
- `pnpm test:ci` — core **1520 passed**; `pnpm format:check`, `pnpm lint`, `pnpm type-check` clean; `pnpm precommit` on every commit
- Playwright `eat-visualization` against the real phosphatase bundle — proves the shader **compiles** (string tests cannot) and the marker sampling still holds
- Glyph harness re-rendered after every shader change and inspected: circles and diamonds, filled and predicted, at three sizes over both surfaces
- The self-transfer command documented in `data/eat_demo/README.md` reproduces the shipped `phosphatase_selffilled.parquetbundle` **byte-identically** (84 `ec` + 52 `protein_families` filled, 0 self-sourced)
- `openspec validate --strict` passes

Every new test was watched failing first. For #393 the tests were additionally re-run against the pre-change `classify()` to confirm all five fail there and pass here.

> Note on running the tests: use `apps/protspace/` as the working directory, as CI does. A bare `uv run pytest` at the repo root resolves `configfile` to the root `pyproject.toml`, which defines no `[tool.pytest.ini_options]`, so `apps/prep` loses its `asyncio_mode = "auto"` and 47 async tests fail to collect. Both CI jobs set `working-directory` and are unaffected.

## Related

**#424** (filed, not implemented) — EAT audit mode. Worth knowing why it matters: the manuscript's headline audit result is computed in the space its own Methods section forbids. `03.results.md:37` derives ~860 candidate misannotations from *"50 nearest **UMAP** neighbors"*, while `05.methods.md:73` states that using the 2-D projection rather than the embedding space is *"a correctness requirement rather than a performance choice"*.

Also surfaced and **not** addressed here: the now-reachable no-filter default is the most expensive configuration the command has (~300K queries × ~270K references at Swiss-Prot scale) and runs with no progress output, since `protlabel` has no tqdm anywhere. Before this PR the same command failed instantly, so no user could reach that path.

https://claude.ai/code/session_016tSUUWsGomA4BFoVFGvSwj


tsenoner:
Companion: #427 carries the #380 reliability-filter work, stacked on #416.

tsenoner:
## Correction: the outline was too thin, now fixed

@tsenoner spotted that the outline was invisible in the app. He was right, and the original commit here was wrong.

Measuring the outline purely in device pixels dropped it to **1 px at every point size** — roughly an **18× reduction** at the default `gl_PointSize` of 240 (18 px → 1 px). The rim was effectively gone while exploring.

| pointSize | radius (px) | old band (px) | device-pixel-only (px) |
|---:|---:|---:|---:|
| 48 | 24 | 3.6 | 1.0 |
| 240 | 120 | 18.0 | 1.0 |
| 512 | 256 | 38.4 | 1.0 |

The "0.35 px at pointSize 48, 1.13 px at 512" figures in the first commit message and in this PR description were **wrong** — out by an order of magnitude and inverted in direction. The print concern behind them is real, but it is a *floor* problem, not a reason to drop the fraction.

**Fixed** in `3604f5e`: the width is now `max()` of both terms — a fraction of the sprite radius (the same 0.15 as before, so on-screen weight matches what shipped previously) and a device-pixel floor so it survives small sprites and export scale. The rim is on predicted rings too, still budget-capped so it cannot eat the annulus.

### Verified by looking at it

Added `apps/web/tests/glyph-outline.spec.ts`: compiles the **real** point shader against a bare WebGL2 context and draws filled dots beside predicted rings, three point sizes, over both surfaces. This change went wrong precisely because it was reasoned about rather than looked at. It also fails if the shader stops compiling or linking, which the string-matching tests cannot catch. Screenshot output is gitignored.

### Two things the render exposed — both pre-existing, not from this PR

- **Dark categories get no visible rim.** The outline is `finalColor * 0.5`, a *relative* darken, so on an already-dark colour there is nothing to darken. Hue-relative contrast is the weakness a theme-resolved stroke uniform would fix.
- **On a dark surface the rim darkens *toward* the background**, acting as a vignette that shrinks the dot rather than outlining it.

Both argue for an explicit stroke colour rather than a fill multiply. That is a wider change (uniform plumbing through 6 files plus the legend renderers) and belongs in its own PR.

PR #430, opened 2026-08-11
fix(embed): an incomplete embedding must not exit 0

jcoludar:
## What happened

While preparing datasets for a paper we ran `protspace embed` on 832 sequences and got this:

```
Embedding: 100%|██████████| 832/832 [01:05<00:00, 12.74seq/s]
Done. Embedded 0 / 832 sequences.
Failed batches: 1 (rerun to retry)
$ echo $?
0
```

Progress bar at **100%**, a 6 KB `.h5` on disk, exit status **0**, and **zero** sequences embedded.

It then happened a second time in a more dangerous form. On the retry, one batch took an **HTTP 403 from Cloudflare** — ordinary rate limiting, provoked simply by having two protspace jobs running against Biocentral at once — and the result was:

| run | batch size | outcome | exit |
|---|---|---|---|
| 1 | 1000 | `Loading of embeddings failed before export!` → **0 / 832**, 6 KB `.h5` | **0** |
| 2 | 100 | **HTTP 403** on one batch → **748 / 832** (89.9 %) | **0** |
| 3 | 100 | 403 again → **776 / 832** | **0** |
| 4 | 40 | 403 again → **804 / 832** | **0** |

**Run 2 onwards is the case that matters.** A 90 %-complete embedding is much more dangerous than an empty one: an empty run at least leaves a conspicuously tiny file, whereas a truncated `.h5` projects, bundles and scores completely normally. UMAP runs on the subset, `--stats` returns plausible silhouettes, and every downstream number is quietly computed on a dataset missing 10 % of its proteins. Nothing in the pipeline re-checks the count.

We only noticed because we had written an external checker (dataset count + id-set against the source FASTA). Without it a truncated file would have gone straight into an experiment.

## Why the failure was invisible

Four separate paths in `embed_sequences` swallowed a failure. We verified each by driving a stubbed `BiocentralAPI` through the real CLI:

| scenario | bar | exit | error shown? |
|---|---|---|---|
| A exception inside the batch loop | 0 % | 0 | yes |
| B `result is None` | **100 %** | 0 | yes |
| C non-`None` result carrying an **empty dict** | **100 %** | 0 | **nothing at all** |
| D API returns a **subset** (`missing_reps`) | **100 %** | 0 | warning only |

- **B** incremented `failed_batches` and then fell through to the shared `pbar.update(batch_protein_count)`, which sat *outside* the `if/else`, so the bar advanced fully anyway.
- **C** — `if emb_dict:` had no `else`. A non-`None` result with an empty dict incremented no counter, logged nothing, and still advanced the bar. Completely silent.
- **D** — `missing_reps` was a `logger.warning` with **no counter**, and the bar still advanced by the full batch size. This is the partial-truncation case above.

Separately, the empty 6 KB `.h5` is not written by `embed_sequences` at all — it is created by `cli/embed.py` opening the file to write the `model_name` attribute after the failure, which then prints an affirmative `Saved: … (model_name=…)`.

## What this changes

**`data/embedding/biocentral.py`**

- Collapse B and C into a single `if not emb_dict:` branch that counts and logs.
- Treat a short response (`missing_reps`) as a partial failure with a counter, not a bare warning.
- **Advance the bar by `len(expanded)` — what was actually written.** The bar position, `total_embedded`, and the file contents can no longer drift apart. This is the change that makes the others hard to regress.
- Raise `ValueError` at the end when `total_embedded < len(remaining)`, with distinct messages for "nothing embedded" and "partially embedded".

`ValueError`, not `RuntimeError`: `cli/prepare.py` already catches `(FileNotFoundError, ValueError)` and turns it into a clean `ERROR: <msg>` + `typer.Exit(1)`. A `RuntimeError` escapes that handler as a raw traceback, and `RuntimeError` appears nowhere else in the package.

**`cli/embed.py`**

- Add the stage-failure handler that `prepare` has and `embed` lacked, so the message is a message rather than a traceback.
- Because the `h5py.File(h5_path, "a")` call now sits behind it, the empty `.h5` and the false `Saved:` line are no longer produced on failure.

Partial output is still written and resume-by-key is untouched, so a rerun embeds only what is missing.

## Why this matters beyond the CLI

PR #280 taught the prep service to classify Biocentral outages, and its spec keys on the `embed` step **exiting non-zero**. A partial embedding exits 0 today, so the prep service never classifies it, proceeds with a truncated `.h5`, and the web app ships a truncated bundle. There is already a production consumer depending on the exit status this bug fails to set.

## Verification

- Full suite: **806 passed, 2 skipped**.
- Targeted: `test_biocentral_embedder.py`, `test_backend_switch.py`, `test_local_embedder.py` — 52 passed, 2 skipped.
- Acceptance test driving all five paths against a stubbed API:

| scenario | before | after |
|---|---|---|
| A exception | bar 0 %, exit 0 | `ValueError`, 0/20 written |
| B returns `None` | bar 100 %, exit 0 | `ValueError`, 0/20 |
| C empty dict | bar 100 %, exit 0, silent | `ValueError`, 0/20 |
| D subset | bar 100 %, exit 0, warn-only | `ValueError`, **8/20 written, bar 40 %** |
| E success | — | no raise, 20/20, bar 100 % |

## One judgement call worth your view

With "all or fail", a sequence the server can *never* embed makes every rerun exit 1, and `prepare` could never complete. We deliberately did **not** paper over this by downgrading the partial case back to a warning — that is the bug. The honest escape is an explicit opt-in (`--allow-partial`, or a documented threshold), which felt like your call rather than ours, so it is not in this PR.

Two smaller things noticed nearby and left alone: `local.py` has the same bar-advance-on-skip shape in its OOM branch, and its final `if not load_existing_ids(h5_path)` guard is masked by resume (a rerun that embeds nothing new still finds prior-run keys and reads as success). Happy to follow up separately if useful.

### DIFF
--- a/apps/protspace/src/protspace/cli/embed.py
+++ b/apps/protspace/src/protspace/cli/embed.py
@@ -98,14 +98,22 @@ def embed(
         h5_path = output / f"{model_name}.h5"
 
         logger.info(f"Embedding with {model_name} ({backend.value}) → {h5_path}")
-        embed_sequences(
-            sequences,
-            resolve(model_name),
-            h5_path,
-            embed_config=embed_config,
-        )
+        try:
+            embed_sequences(
+                sequences,
+                resolve(model_name),
+                h5_path,
+                embed_config=embed_config,
+            )
+        except (FileNotFoundError, ValueError) as e:
+            # Mirrors the stage-failure handler in cli/prepare.py. Without it an incomplete
+            # embedding surfaces as a raw traceback instead of a message.
+            logger.error(str(e))
+            raise typer.Exit(1) from e
 
-        # Write model_name attr
+        # Write model_name attr. Only reached on success — on failure this h5py.File(..., "a")
+        # call is what fabricated the empty ~6 KB .h5, which together with the affirmative
+        # "Saved:" line below made a total failure look like a finished run.
         with h5py.File(h5_path, "a") as f:
             f.attrs["model_name"] = model_name
 
--- a/apps/protspace/src/protspace/cli/embed.py
+++ b/apps/protspace/src/protspace/cli/embed.py
@@ -94,6 +94,8 @@ def embed(
         )
         resolve = resolve_embedder
 
+    failed_models: list[str] = []
+
     for model_name in embedder:
         h5_path = output / f"{model_name}.h5"
 
@@ -106,15 +108,26 @@ def embed(
                 embed_config=embed_config,
             )
         except (FileNotFoundError, ValueError) as e:
-            # Mirrors the stage-failure handler in cli/prepare.py. Without it an incomplete
-            # embedding surfaces as a raw traceback instead of a message.
+            # Same stage-failure shape as cli/prepare.py. The models are independent
+            # (one .h5 each), so carry on and report every failure at the end rather
+            # than abandoning the models that have not been tried yet.
             logger.error(str(e))
-            raise typer.Exit(1) from e
+            failed_models.append(model_name)
+            continue
 
-        # Write model_name attr. Only reached on success — on failure this h5py.File(..., "a")
-        # call is what fabricated the empty ~6 KB .h5, which together with the affirmative
-        # "Saved:" line below made a total failure look like a finished run.
+        # Write model_name attr. Skipped on failure, so a run that embedded nothing
+        # no longer leaves a stamped .h5 and an affirmative "Saved:" line behind.
         with h5py.File(h5_path, "a") as f:
             f.attrs["model_name"] = model_name
 
         typer.echo(f"Saved: {h5_path} (model_name={model_name})")
+
+    if failed_models:
+        if len(embedder) > 1:
+            logger.error(
+                "Embedding failed for %d of %d model(s): %s",
+                len(failed_models),
+                len(embedder),
+                ", ".join(failed_models),
+            )
+        raise typer.Exit(1)
--- a/apps/protspace/src/protspace/data/embedding/biocentral.py
+++ b/apps/protspace/src/protspace/data/embedding/biocentral.py
@@ -143,6 +143,9 @@ def embed_sequences(
     Supports deduplication, length-sorting, batching, resume (skips IDs
     already present in *h5_path*), and a tqdm progress bar.
 
+    Raises ``ValueError`` if any requested sequence could not be embedded; the
+    partial HDF5 is kept so a rerun only embeds what is missing.
+
     Returns the path to the completed HDF5 file.
     """
     cfg = embed_config or EmbedConfig()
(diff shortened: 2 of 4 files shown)


tsenoner:
Reviewed this and reproduced the whole report — the diagnosis is right and the bar fix is the real root-cause fix. I drove the pre-PR code through the CLI with a stubbed API and got your table back exactly, including the 6 KB file:

| failure mode | exit | bar | written | `.h5` | `Saved:` |
|---|---|---|---|---|---|
| exception in batch | 0 | 100% | 0/20 | 6144 B | yes |
| `result is None` | 0 | 100% | 0/20 | 6144 B | yes |
| empty dict | 0 | 100% | 0/20 | 6144 B | yes |
| short response | 0 | 100% | 8/20 | 14464 B | yes |

The prep-service argument is stronger than you wrote it, too: `pipeline.py:152` guards with `if not h5_files`, and the fabricated 6 KB file **defeats that guard**. So a partial embed exited 0, passed the guard, and shipped a truncated bundle.

I've pushed three commits onto this branch rather than leaving them as comments, because the first one fixes a bug the PR itself introduces — happy to drop or rework any of them.

**dde4f4d2 — the completeness gate counted intent, not disk.** `total_embedded += len(expanded)` counts what was handed to `save_embeddings`, which silently skips IDs already present (`if protein_id not in f`). h5py also turns an ID containing `/` into a *group*, so two requested IDs can collapse into one dataset. A FASTA with `>A/B` and `>A`:

```
Done. Embedded 2 / 2 sequences.     ← exit 0
h5 top-level keys: ['A']            ← one dataset, neither ID retrievable
```

Now gated on `set(remaining) - load_existing_ids(h5_path)` — the same disk-based check `local.py:358` already uses. That removes `total_embedded` entirely, so the two raise branches share one detail string.

**9a6b0aaf — `raise typer.Exit(1)` abandoned the remaining models.** `-e prot_t5 -e esm2_8m -e esm2_150m` with model 1 failing attempted only model 1; pre-PR all three ran. The models are independent (one `.h5` each), so it now collects failures, continues, and exits 1 once at the end.

**c7be1383 — the acceptance test.** The PR body describes a 5-scenario test but the diff had none. Added 9 covering all five paths plus bar accounting, the disk-vs-counter gate, rerun convergence, and multi-model continuation. `856 passed, 2 skipped`.

Same commits also trimmed the completeness path: three `pbar.set_postfix` calls folded into one per iteration, and `Done.`/`Output:` moved below the gate — it was printing `Done. Embedded 0 / 832` immediately before exiting 1, which is the same affirmative-on-failure output the PR exists to remove.

---

On your judgement call — agreed, `--allow-partial` is the honest escape and shouldn't be smuggled in here. Two things worth a follow-up issue:

1. **`--backend local` still has this exact bug.** `local.py:292-303` drops sequences over `max_length=2000` and `local.py:337` drops OOM'd ones with only a `logger.warning`; the sole guard fires when the file is *completely* empty. And `resolve_default_backend()` returns `"local"` on Colab GPU — so the unguarded backend is the default for exactly the audience `BIOCENTRAL_UNAVAILABLE` routes to Colab. Deliberately not touched here: the `max_length` drop is intentional, so a shared check needs a policy call on "deliberately excluded" vs "failed".
2. **Nothing downstream cross-checks.** `load_h5` never sees the FASTA, and `pipeline._validate_headers` returns `embedding_sets[0].headers` unchecked for the single-embedder case. A truncated `.h5` from an older version, an interrupted run, or a Ctrl-C still projects and bundles normally — this PR closes the door on one producer while the consumer stays open.

Also noted, not acted on: a dead server still grinds through every batch before the raise fires (~20 s per failing batch × 570 batches ≈ 3.3 h on a Swiss-Prot-scale run). A break after N consecutive failed batches would fall through to the same message.

PR #435, opened 2026-08-11
docs(openspec): record the filter-query semantics #416 shipped

tsenoner:
Closes the OpenSpec debt left by #416.

#416 was ~1,600 lines across 22 files. It **redefined `NOT`** for every saved query, added the `ANY_VALUE` presence sentinel, added the inclusive `gte`/`lte` operators, and rewrote the EAT reliability filter — and it shipped with no `openspec/changes/` entry. `AGENTS.md` makes `openspec/specs/` the source of truth for current behaviour, and `openspec/specs/` carried **no filter-query capability at all**, so nothing outside the source recorded the one rule every saved query depends on.

This adds the change retroactively and archives it, so `openspec/specs/filter-query-semantics/` becomes true.

## What the spec now states

Eight requirements, each with scenarios, all derived from the code on `main`:

- **`NOT` is scoped to proteins that carry a value** — `difference(proteinsWithAnyValue(annotations), matches)`, not a set complement. That is what makes `NOT (is N/A)` mean exactly "has any value", and what makes the hand-written `AND NOT (is N/A)` guard redundant. For a group the scope is "carries a value for **at least one** of the annotations", not all.
- **Presence sentinels** — `NA_VALUE` / `ANY_VALUE` as exact complements, carried in `values` on the categorical side and `presence` on the numeric side, evaluated identically.
- **Presence unions with the comparison** — `>= 0.5` plus an N/A chip reads "at least 0.5, or no value at all". A null is matched *only* by an explicit N/A chip; no comparison operator ever matches one.
- **`ANY_VALUE` is exclusive** — selecting it replaces the selection and locks the rest out, enforced in the picker, the categorical select handler and the numeric chip handler.
- **One readiness rule** — a numeric condition is configured when it has its operator's bounds **or** a presence chip; an unready condition stays a match-all no-op.
- **Inclusive `gte`/`lte`**, and switching operator nulls the bound the new operator does not use.
- **The EAT reliability filter** is `EAT_confidence >= x` + N/A chip. It was `NOT(EAT_confidence < x)`, which kept curated points only as a side effect of the old complement — under the new `NOT` it would have hidden every one of them.
- **Previews agree with results** — the picker's `NOT` count is the has-a-value slice minus the carriers *inside that slice*, cross-checked against `evaluateQuery`; `countNumericMatches` walks `protein_ids` so a sparse column can't undershoot.

## What it deliberately does not state

The findings from the #416 review that were reported and left unfixed are recorded as **Non-Goals / Open Questions in `design.md`**, not as requirements — a spec should not bless a known gap:

- the numeric row's live count ignores `logicalOp`, so a `NOT`'d numeric row previews the un-negated count;
- reliability-condition orphaning, fixed separately in #427 by keying on a declared owner rather than the condition's shape;
- `presence?: string[]` as a 3-state enum modelled as an unbounded array;
- the Kleene / SQL-NULL refactor that would make `NOT` fall out as `defined ∖ matched` and delete `collectAnnotations` + `proteinsWithAnyValue`;
- the `<option>` list being the one un-type-checked half of the operator table.

`tasks.md` records every task against the commit that actually did it (`e27c7be8`, `1aedc853`, `a3b65b2a`, `082b8b75`), since the work is already on `main`.

The five stale `NOT(EAT_confidence < x)` requirements in `add-eat-visualization` were already corrected in #416 and are untouched here.

## Verification

- `openspec validate --all --strict` — **16 passed, 0 failed**, including the new `spec/filter-query-semantics`
- `pnpm format:check` clean
- Docs-only: no source, no tests, no behaviour change.

https://claude.ai/code/session_01HGVhbzHK79uUDaJPhTDu9c

### DIFF
--- a/openspec/changes/archive/2026-08-11-add-filter-presence-semantics/.openspec.yaml
+++ b/openspec/changes/archive/2026-08-11-add-filter-presence-semantics/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-11
--- a/openspec/changes/archive/2026-08-11-add-filter-presence-semantics/design.md
+++ b/openspec/changes/archive/2026-08-11-add-filter-presence-semantics/design.md
@@ -1,0 +1,138 @@
+## Context
+
+The query builder evaluates a `FilterQuery` — a list of conditions and groups — into a `Set<number>`
+of protein indices (`query-evaluate.ts`). Conditions come in two kinds that had drifted apart:
+categorical conditions selected values from a list that already contained the `NA_VALUE` sentinel,
+while numeric conditions had only an operator and one or two bounds and no concept of a missing
+value at all.
+
+`NOT` lived at the combining level, applied as `allIndices ∖ itemResult`. Because a protein missing
+a value simply fails every condition, complementing swept all of them in. Two consequences shaped
+this change:
+
+1. Users wrote `NOT (X) AND NOT (is N/A)` by hand, and the second half is the part that is hard to
+   discover — a negation that silently answers with the un-annotated majority looks like a data bug,
+   not a semantics choice.
+2. The EAT reliability slider (`control-bar.setEatConfidenceThreshold`) had been built to exploit the
+   sweep: `NOT(EAT_confidence < x)` kept curated proteins visible only because they are null and the
+   complement returned them. That made the reliability filter's correctness depend on a defect.
+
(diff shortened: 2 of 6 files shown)


tsenoner:
Superseded — merged into `feat/filter-design-system` instead, so the OpenSpec catch-up lands together with #417 rather than separately. #417's own closing OpenSpec change is going onto the same branch.

Nothing here is dropped: every commit is now on `feat/filter-design-system`.

PR #476, opened 2026-09-04
Landing page redesign

peymanvahidi:
I redesigned the landing page, I think now we can showcase the features that we have in ProtSpace better. 

This is the initial idea. check it out and tell me what you think.

I think we should do couple of iterations on adding some other datasets, changing some texts and a bit more polishing. Also it's a bit interactive, which is also sth I like :)


t03i:
Absolutely love the redesign! I would say ship it and iterate over it on main 😍

peymanvahidi:
Works for me!

tsenoner:
Screenshot 2026-09-16 at 18 44 24

peymanvahidi:
- Border lines around the data points are too dark
- Add "shift + hover" option
- add lables for "venom kunitz", "snaclec", "arthropod" families in the landing page.
- consider adding an option to add lable on the plot
  - maybe taking top 40% of the data that have the most density.
- remove the 3 lables -> ProtT5-umap, num proteins, sp demo bundle
- text: Explore protein language model embeddings overlaid with biology — and see what sequence similarity can’t.
- button for docs
- remove the "Projection quality" part
- make the web page cleaner

PR #478, opened 2026-09-10
feat(webgl): add contour and density plots

peymanvahidi:
Now we have a new drop down (next to search field) which you can switch between density plot and contour plot.

some notes:
- we can change the drop down and move it do settings. I just put it here for ease of use, I have no preferences regarding this.
- we have off/auto/on options for each. I was thinking that we can only keep the auto and off. for auto mode when you zoom in after a threshold you'll see the scatterplot without density/contours, which is nice (I guess).
- also take a closer look to contour lines, their colors are mixing with each other. a feature that I like.
- some knobs, like opacity of contour backgrounds, thickness of the lines, etc can be set. I chose these for now.
- Not sure if density plot is sth that we would like to keep. the logic behind is that the colors of the data points which are overlapping are being mixed.

Screenshot 2026-09-10 at 14 41 19

Screenshot 2026-09-10 at 14 40 53


peymanvahidi:
I made the lines thicker and I also suggest to:
- make the default `shape size` to `5`
- preserve the `shape size` when switching annotations. I think it's more convenient

in this pic the lines are thicker and the `shape size` is 5

Screenshot 2026-09-10 at 21 28 22

peymanvahidi:
maybe we can add a feature that when we zoom, the dots gets a bit larger

tsenoner:
TODO: make contours based on the categories and not one merged density contour with overlapping colors.

